PR #2, opened 2020-04-29
Refactoring

### LINKED ISSUE #1: Barebones API spec for explorer/dashboard
Placeholder issue. This is mostly in place already.

avive:
- Added several methods
- Separated types and service to different proto files 
- Added fields to many of the core data types.

Closes #1


proto/spacemesh/types.proto
```diff
@@ -0,0 +1,162 @@
+syntax = "proto3";
+package spacemesh;
+
+message SimpleInt {
```
lrettig, line 4: Do we still need this if we have Amount, below?

proto/spacemesh/types.proto
```diff
@@ -0,0 +1,162 @@
...
+    uint64 value = 1;
+}
+
+message AccountId {
+    bytes address = 1;
+}
+
+message Account {
+    AccountId address = 1;  // account public address
+    bytes public_key = 2;   // nice to include the account full public key
```
lrettig, line 23: I think I'd leave this out, it bloats the data and the API consumer should be agnostic to this

proto/spacemesh/types.proto
```diff
@@ -0,0 +1,162 @@
...
+    PENDING = 4;                // pending processing by STF
+    //PROCESSING = 5;
+    //APPROVED = 6;
+    PROCESSED = 7;              // processed. Results are in the tx receipt.
+}
+
+enum TransactionType {
+    COIN = 0;            // simple coin transaction
+    ATX = 1;             // smesher activation transaction
+    APP = 2;             // smart contract method
```
lrettig: I'd prefer to avoid the word "app" here as it's confusing... can we just call this smart contract, or else "script" or something?

proto/spacemesh/types.proto
```diff
@@ -0,0 +1,162 @@
...
+    INSUFFICIENT_FUNDS = 2;     // rejected pre STF processing by funds check.
+    CONFLICTING = 3;            // rejected pre STF due to conflicting counter.
+    PENDING = 4;                // pending processing by STF
+    //PROCESSING = 5;
+    //APPROVED = 6;
+    PROCESSED = 7;              // processed. Results are in the tx receipt.
+}
+
+enum TransactionType {
+    COIN = 0;            // simple coin transaction
```
lrettig: s/COIN/SIMPLE/ ?

proto/spacemesh/types.proto
```diff
@@ -0,0 +1,162 @@
...
+    INSUFFICIENT_GAS_PROVIDED = 1;      // out of gas
+    INSUFFICIENT_BALANCE = 2;           // failed due to sender insufficient funds
+    EXECUTED = 3;                      // executed w/o error
+}
+
+message TransactionReceipt {
+    TransactionId id = 1;  // the source transaction
+    TransactionResult result = 2;   // tx processing result
+    uint64 gas_used = 3;            // gas units used by the transaction (gas price in tx)
+    Amount fee = 4;                 // transaction fee charged for the transaction
```
lrettig, line 59: How is fee different from gas_used?

proto/spacemesh/types.proto
```diff
@@ -0,0 +1,162 @@
...
+    uint64 gas_price = 2;
+}
+
+// Simple coin transaction
+message Transaction {
+    TransactionType type = 1;
+    TransactionId id = 2; // hash
+    AccountId sender = 3;
+    TransactionFee fee = 4;
+    uint64 timestamp = 5;  // shouldn't this be part of the event envelope?
```
lrettig, line 81: If so, it would just be the time the tx was generated, or signed, which is just informational. I find this confusing and I'd leave it out. You could add a timestamp to the receipt, that would be a little more meaningful.

proto/spacemesh/types.proto
```diff
@@ -0,0 +1,162 @@
...
+    uint64 layer_id = 6;
```
lrettig, line 82: What happens if the same tx ends up in two different layers?

proto/spacemesh/types.proto
```diff
@@ -0,0 +1,162 @@
...
+}
+
+message Reward {
+    uint64 layer = 1;
+    Amount total = 2;
+    Amount layer_reward = 3;
+    uint64 layer_computed = 4; // layer number of the layer when reward was computed
+    // tx_fee = total- layer_reward
+    AccountId coinbase = 5; // account awarded this reward
+    SmesherId smesher = 6;  // it will be nice to always have this in reward events
```
lrettig, line 100: Is there such thing as a `SmesherId`? didn't we say this is going away or something?

proto/spacemesh/api.proto
```diff
@@ -1,166 +1,70 @@
...
-}
+    // Returns node status
+    rpc NodeStatus(google.protobuf.Empty) returns (NodeStatus);
 
-message Block {
-    bytes id = 1;
-    repeated Transaction transactions = 2;
-}
+    // Tell the node to start syncing
+    rpc NodeSyncStart(google.protobuf.Empty) returns (google.protobuf.Empty);
```
lrettig, line 23: Are we okay if these return nothing? Do we want them to return a success/failure status?

proto/spacemesh/api.proto
```diff
@@ -1,166 +1,70 @@
...
 
-message LayerId {
-    bytes id = 1;
-}
+    // Gracefully shut down the node
+    rpc NodeShutdown(google.protobuf.Empty) returns (google.protobuf.Empty);
```
lrettig, line 26: What happens here if the node cannot gracefully shutdown for a long time (per @noamnelke's comments)? Does this succeed immediately but the node doesn't actually shutdown right away? Does it fail? Does a sequence of commands need to be sent to gracefully shutdown?

proto/spacemesh/api.proto
```diff
@@ -1,166 +1,70 @@
...
-message LayerEvent {
-    LayerId layer_id = 1;
-    enum Event {
-        UNDEFINED = 0;
-        APPROVED = 1;
-        CONFIRMED = 2;
-    }
-}
+    // Returns mesh genesis time as unix epoch time
+    rpc MeshGenesisTime(google.protobuf.Empty) returns (SimpleInt);
```
lrettig, line 31: Might be good to add an explicit time type

proto/spacemesh/api.proto
```diff
@@ -1,166 +1,70 @@
...
+    // Returns number of layers per epoch (a network constant)
+    rpc MeshEpochNumLayers(google.protobuf.Empty) returns (SimpleInt);
 
-service SpacemeshService {
-    // SCALAR
+    ////////// Node streaming data facade
 
-    // Test facade
+    // Sync status events
+    rpc NodeSyncStatuses(google.protobuf.Empty) returns (stream NodeSyncStatus);
```
lrettig, line 48: Can we stick to the naming convention I proposed? `NodeStreamSyncStatus` sounds more intuitive to me. I'd like to avoid plurals (that's ambiguous) - it's clear from the `Stream` that this is a stream.

proto/spacemesh/api.proto
```diff
@@ -1,166 +1,70 @@
...
-    // Returns node status object
-    rpc NodeStatus(google.protobuf.Empty) returns (NodeStatus);
-    // Tell the node to start sync
-    rpc NodeSyncStart(google.protobuf.Empty) returns (google.protobuf.Empty);
+    // Receipts - emitted after tx was processed by STF (or rejected before STF)
+    rpc MeshTransactionReceipts(google.protobuf.Empty) returns (stream TransactionReceipt);
 
-    // Mesh facade
+    // App Events - emitted by app methods impl code trigged by an app transaction
+    rpc MeshAppEvents(google.protobuf.Empty) returns (stream AppEvent);
```
lrettig, line 59: Think we should leave this out from the current API version - this is part of the VM - can we bump to 0.0.2?

proto/spacemesh/api.proto
```diff
@@ -1,166 +1,70 @@
...
+    //////////////// Mesh streaming data facade
 
-    // Returns the version of the node software as a semver string
-    rpc NodeVersion(google.protobuf.Empty) returns (SimpleMessage);
-    // Returns node status object
-    rpc NodeStatus(google.protobuf.Empty) returns (NodeStatus);
-    // Tell the node to start sync
-    rpc NodeSyncStart(google.protobuf.Empty) returns (google.protobuf.Empty);
+    // Receipts - emitted after tx was processed by STF (or rejected before STF)
+    rpc MeshTransactionReceipts(google.protobuf.Empty) returns (stream TransactionReceipt);
```
lrettig, line 56: Same - this is part of the VM - let's bump to a later API version

proto/spacemesh/api.proto
```diff
@@ -1,166 +1,70 @@
...
-    // Mesh stream facade
-
-    // Account events
-    rpc MeshStreamAccount(google.protobuf.Empty) returns (stream AccountEvent);
-    // Reward events
-    rpc MeshStreamReward(google.protobuf.Empty) returns (stream Reward);
-    // Layer events
-    rpc MeshStreamLayer(google.protobuf.Empty) returns (stream Layer);
-    // Layer state change events
-    rpc MeshStreamLayerState(google.protobuf.Empty) returns (stream SimpleInt);
```
lrettig, line 165: What happened to this stream?

proto/spacemesh/api.proto
```diff
@@ -1,166 +1,70 @@
...
-    // Returns current layer
-    rpc MeshCurrentLayer(google.protobuf.Empty) returns (SimpleInt);
-    // Returns current epoch
-    rpc MeshCurrentEpoch(google.protobuf.Empty) returns (SimpleInt);
-    // Returns network ID
-    rpc MeshNetId(google.protobuf.Empty) returns (SimpleInt);
-    // Returns number of layers per epoch
-    rpc MeshEpochNumLayers(google.protobuf.Empty) returns (SimpleInt);
+    // Accounts
+    rpc MeshAccounts(google.protobuf.Empty) returns (stream Account);
```
lrettig, line 62: If we replace `AccountEvent` with `Account`, it's no longer possible to tell directly from the stream what changed. It becomes contextual/stateful, i.e., you have to compare with the previous state to see what changed. Maybe this is okay, but we should pick one scheme and stick with it. Might be worth discussing with the explorer team. I'll open an issue.

proto/spacemesh/api.proto
noamnelke: You’re referring to the “very graceful shutdown” that I described? If yes, this isn’t really about shutting down... A more fitting name is “graceful retreat from mining”, which may or may not be followed by a node shutdown.

Since the procedure I described would take 1-2 epochs to complete, it makes no sense to keep a request pending this long.

If you’re referring to the regular node shutdown, then I think it makes sense to return a success or failure (e.g. couldn’t dump recovery data to disk), but this would be very difficult to implement.

lrettig: Maybe we need two endpoints, a "shutdown now" endpoint and a "take your time and shutdown gracefully" endpoint :)

noamnelke: I don’t think it makes sense to order the node to shut down 2 weeks in advance...

The `StopMining` endpoint should only stop creating ATXs and it should be the caller’s responsibility to monitor and shut down the node when the time is right.

We may also want to make a `shutdown` endpoint that fails if the miner is eligible, unless a `force` flag is also passed.

proto/spacemesh/types.proto
avive: Yes - for methods int params which are not coin amounts.

proto/spacemesh/types.proto
avive: Do you mean the public key field? It is for the explorer. We need to display these in it.

proto/spacemesh/types.proto
avive: Let's leave it for now as this is the canonical name in SVM - if we decide to change App to something else then we'll do global replace which will impact many more places. For consistency sake, we should use existing names until we decide to change them and then change everywhere.

proto/spacemesh/types.proto
avive: I don't understand this comment - can you please expand a bit?

proto/spacemesh/types.proto
avive: gas_used is in gas units. Fee is the fee in smidge which is equal to gas_used * gas_price set by transaction caller.

proto/spacemesh/api.proto
avive: Yes, I removed all Streams from the streaming methods because I thought they are too long and it is obvious by looking at the method signature that it returns a stream of messages. I think it is good to use plurals for streaming methods to indicate that they return a stream of messages. Right now - all methods which return streams have plural names.

proto/spacemesh/api.proto
avive: We are going to have smart wallets in 0.2 and events and receipts are quite isolated but we can deal with these in 0.0.2 as you describe as long as we will provide them for this project - the explorer version spec includes full apps events, receipts and state...

proto/spacemesh/api.proto
avive: Same response as above the previous comment :-)

proto/spacemesh/api.proto
avive: I don't think we need it. I think it is good design to just have stream of entities - layers, blocks, transactions and just have the state of these data objects up-to-date so it is the client responsibility to update its stored data based on the last entity it got on a stream and can maintain state like that. So, the idea is that instead sending a this-object-changed-state in a method - we just send the object on its standard stream with the updated state and the client inspects it and updates itself... I think this is a good pattern - I know it was me who suggested this method before...

proto/spacemesh/api.proto
avive: See my above comment - I think we can do without any Events per-se and it is a simpler design which is satisfactory. I don't think there's a way to get around clients having to look for state changes in data objects in stream - for example, transactions, layers, blocks can all change and be sent multiple times... It is actually good the node api is very low-level and doesn't indicate state changes just RAW data and the responsibility to understand state changes moves up to the data workers in client backends... So this is the scheme I think we should stick to and mock for now. We can always add data items state events methods later. This will also greatly simplify logic in the node as it doesn't need to know when a state that requires to raise an event occurs and deal with it...  This I think also the main way we handle reogrs elegantly. In a reorg, the layers, blocks and txs are just sent again over the standard entities streams and the client updates its date based on the current fields in these entities....

proto/spacemesh/types.proto
lrettig: TODO: remove this, since a tx can be in multiple layers

noamnelke: you could return the layer the tx was applied in, as we do today (so the first layer in which it appears in a valid block, that's in concensus).

proto/spacemesh/types.proto
lrettig: Strictly speaking, according to the guidelines I read for API design, each RPC method should have its own unique parameters. They shouldn't be shared, to avoid issues, so there should be no "generic" types. Anyway I think it's okay for now.

proto/spacemesh/types.proto
lrettig: Why should the explorer display public key? Why should it display keys at all? It only needs to display addresses. I've never seen an explorer that displays keys :)

lrettig: What I'm trying to say is that I think the Account should be agnostic to how we implement keypairs. Maybe we decide to change the keys in the future (quantum resistance? multiple types of keys?).

proto/spacemesh/types.proto
lrettig: Okay. Let's discuss elsewhere. I still find this very confusing, though.

proto/spacemesh/types.proto
lrettig: Can we call this a "SIMPLE" transaction rather than a "COIN" transaction?

proto/spacemesh/types.proto
lrettig: I think it might be more intuitive if, instead of the receipt containing `gas_used` and `fee`, it instead contained `gas_used` and `gas_price`, like the tx itself. The math is the same.

proto/spacemesh/api.proto
lrettig: Moved to #10

proto/spacemesh/types.proto
lrettig: We can discuss further in #8

proto/spacemesh/api.proto
lrettig: > I think it is good to use plurals for streaming methods to indicate that they return a stream of messages

I disagree here :) I think it's ambiguous. But anyway we're going to refactor into different services so it should be obvious from the context (service name) which endpoints are streams.

proto/spacemesh/api.proto
lrettig: Resolved in #5

PR #3, opened 2020-04-29
CI

lrettig:
Based on https://github.com/bufbuild/buf-example/


lrettig:
Added travis and circleci config as well but it seems to work fine with github-actions for now.

PR #16, opened 2020-05-08
Smeshing methods (v0.2)

### LINKED ISSUE #13: API v2 including STF-related things
Placeholder for ongoing work

avive:
Closes #13


lrettig:
This is based on the wrong branch. Could you please rebase on the v0.2 branch? (https://github.com/spacemeshos/api/commits/v0.2)

avive:
should be ready for review now.

proto/spacemesh/types.proto
```diff
@@ -128,14 +202,33 @@ message Layer {
...
-    uint64 verified_layer = 7;
+    uint64 connected_peers = 1;       // number of connected neighbors
+    bool   is_synced = 2;             // Is meshed synced
+    uint64 synced_layer = 3;          // the last layer node has synced
+    uint64 top_layer = 4;             // top layer is the tip
+    uint64 verified_layer = 5;        // the last layer node has verified
+}
+
+// Indicates a change in node status
+message NodeStatusEvent {
```
lrettig, line 196: You replaced `NodeSyncStatus` with this type, but this type isn't used in any of the services. Also, it's missing

```suggestion
message NodeStatusEvent {
    NodeStatusEvent status = 1;
```

proto/spacemesh/smesher_types.proto
```diff
@@ -0,0 +1,44 @@
...
+// User provided post data
+message CommitmentUserData {
+    string path = 1;                // Path to post file(s) on a volume
+    uint64 commitment_size = 2;     // User specified commitment size in bytes
+    uint32 processor_id = 3;        // Processor to be used for smeshing. e.g. cpu, gpu0, gpu1, etc...
+    bool   append = 4;              // Append to existing file if it exists. Otherwise - overwrite
+    bool   throttle = 5;            // Throttle down init phase computations while user is interactive on system
+}
+
+message CommitmentProcessor {
```
lrettig, line 13: Please add description

proto/spacemesh/smesher_types.proto
```diff
@@ -0,0 +1,44 @@
...
+    uint32 id = 1;                  // Numerical identifier - unique per system
+    string processor_name = 2;      // Proc info, e.g. Nvidia `Geforce RTX`, `AMD Radeon`, `Intel Genuine i7 CPU`
+    string api = 3;                 // Pprocessing api to be used with the processor. e.g. `Vulkan`, `Cuda` or `x86-64`
+}
+
+message CommitmentProcessors {
```
lrettig, line 19: Please add description

proto/spacemesh/smesher_types.proto
```diff
@@ -0,0 +1,44 @@
...
+    repeated CommitmentProcessor processors = 1;
+}
+
+message DeleteCommitment {
```
lrettig, line 23: Please add description

proto/spacemesh/smesher.proto
```diff
@@ -0,0 +1,51 @@
...
+    // Set a min gas units for including txs in blocks by this smesher
+    rpc SetMinGas(SimpleInt) returns (BoolResult);
+
+    /////// PoSpace commitments management
+
+    // Returns post commitment data from the node
+    rpc CommitmentData(google.protobuf.Empty) returns (CommitmentData);
+
+    // Returns a list of available commitment init processors (including CPU as fallback)
+    rpc CommitmentProcessor(google.protobuf.Empty) returns (CommitmentProcessors);
```
lrettig, line 40:
```suggestion
    rpc CommitmentProcessors(google.protobuf.Empty) returns (CommitmentProcessors);
```

proto/spacemesh/smesher_types.proto
```diff
@@ -0,0 +1,44 @@
+syntax = "proto3";
+package spacemesh;
+
+// User provided post data
+message CommitmentUserData {
+    string path = 1;                // Path to post file(s) on a volume
+    uint64 commitment_size = 2;     // User specified commitment size in bytes
+    uint32 processor_id = 3;        // Processor to be used for smeshing. e.g. cpu, gpu0, gpu1, etc...
+    bool   append = 4;              // Append to existing file if it exists. Otherwise - overwrite
+    bool   throttle = 5;            // Throttle down init phase computations while user is interactive on system
```
lrettig, line 10: Should this be bool? Why not make it, e.g., a percentage (max. pct. of CPU to utilize), or a 1-10 scale, or number of cores?

proto/spacemesh/smesher_types.proto
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// Proof of space commitment data
+message CommitmentData {
+    enum FileStatus {
+        NOT_FOUND = 0;  // Commitment file was not initialized (in this or in previous node sessions)
+        COMPLETE = 1;   // Complete commitment file is available and verified
+        PARTIAL = 2;    // Init commitment file is partially written and init can be continued but is not in progress
+    }
+    bool init_in_progress = 2;  // True if init process is currently in progress
```
lrettig, line 34:
```suggestion
    FileStatus file_status = 1;
    bool init_in_progress = 2;  // True if init process is currently in progress
```

proto/spacemesh/smesher_types.proto
```diff
@@ -0,0 +1,44 @@
...
+}
+
+message DeleteCommitment {
+    bool delete_commitment = 1;
+}
+
+// Proof of space commitment data
+message CommitmentData {
+    enum FileStatus {
+        NOT_FOUND = 0;  // Commitment file was not initialized (in this or in previous node sessions)
```
lrettig, line 30: Please add `UNDEFINED = 0` as first element in enum (required by protobuf)

proto/spacemesh/smesher_types.proto
```diff
@@ -0,0 +1,44 @@
...
+        COMPLETE = 1;   // Complete commitment file is available and verified
+        PARTIAL = 2;    // Init commitment file is partially written and init can be continued but is not in progress
+    }
+    bool init_in_progress = 2;  // True if init process is currently in progress
+    string path = 3;            // Commitment file path on a system volume
+    uint64 bytes_written = 4;   // Bytes written to the commitment file (actual file size)
+    uint64 commitment_size = 5; // User provided commitment size in bytes
+    string error_message = 6;   // Last init process error message
+    enum Errors {
+        FILE_NOT_FOUND = 0;     // Post commitment file not found in expected path
```
lrettig, line 40: Please add `UNDEFINED = 0` as first element in enum (required by protobuf)

proto/spacemesh/smesher_types.proto
```diff
@@ -0,0 +1,44 @@
...
+        READ_ERROR = 1;         // Failure to read commitment file
+        WRITE_ERROR = 2;        // Failure to write to commitment file
+    }
```
lrettig, line 43:
```suggestion
    }
    ErrorType error_type = 7;
```

proto/spacemesh/smesher_types.proto
```diff
@@ -0,0 +1,44 @@
...
+        NOT_FOUND = 0;  // Commitment file was not initialized (in this or in previous node sessions)
+        COMPLETE = 1;   // Complete commitment file is available and verified
+        PARTIAL = 2;    // Init commitment file is partially written and init can be continued but is not in progress
+    }
+    bool init_in_progress = 2;  // True if init process is currently in progress
+    string path = 3;            // Commitment file path on a system volume
+    uint64 bytes_written = 4;   // Bytes written to the commitment file (actual file size)
+    uint64 commitment_size = 5; // User provided commitment size in bytes
+    string error_message = 6;   // Last init process error message
+    enum Errors {
```
lrettig, line 39:
```suggestion
    enum ErrorType {
```

proto/spacemesh/smesher_types.proto
avive: done

proto/spacemesh/smesher_types.proto
avive: done

proto/spacemesh/smesher_types.proto
avive: done. Changed to StopSmeshingParams.

proto/spacemesh/smesher_types.proto
avive: yes it should. It is on or off and the actual throttling happens in a lower level dynamically based on user interaction with the system per the current gpu-post init design.

proto/spacemesh/smesher_types.proto
avive: Are you 100% sure about this. The field in the message can only have one of these 3 possible statues and should never return UNDEFINED. Looking at the protobuf docs - they show enums without UNDEFINED = 0 as the first element... where in the docs do you see this is required?

proto/spacemesh/smesher_types.proto
avive: See comment above...

proto/spacemesh/smesher_types.proto
lrettig: This comes from the [protobuf style guide](https://developers.google.com/protocol-buffers/docs/style#enums):

> Each enum value should end with a semicolon, not a comma. Prefer prefixing enum values instead of surrounding them in an enclosing message. The zero value enum should have the suffix UNSPECIFIED.

See also https://buf.build/docs/lint-checkers#enum_zero_value_suffix.

As I said on Slack, I think the most important thing is that we pick a convention and stick to it. I'm inclined to stick to these best practice conventions unless we have a compelling reason to deviate.

Let's move this thread to #17.

proto/spacemesh/types.proto
```diff
@@ -24,14 +28,21 @@ message Account {
...
-    uint64 layer = 1;
-    Amount total = 2;
-    Amount layer_reward = 3;
-    uint64 layer_computed = 4; // layer number of the layer when reward was computed
-    // tx_fee = total - layer_reward
-    AccountId coinbase = 5; // account awarded this reward
-    SmesherId smesher = 6;  // it will be nice to always have this in reward events
+message AccountIdWithOffset {
+    AccountId account_id = 1;
+    uint64 offset = 2;
```
lrettig, line 33: Doesn't having an offset imply persistent state in the backend? Can we get away with this without adding state/complexity to go-spacemesh?

proto/spacemesh/types.proto
```diff
@@ -77,25 +71,45 @@ message TransactionReceipt {
         INSUFFICIENT_GAS = 5;       // out of gas
         INSUFFICIENT_FUNDS = 6;     // failed due to sender's insufficient funds
     }
-    TransactionResult result = 2;
+    TransactionResult result = 2;   // tx processing result
     uint64 gas_used = 3;            // gas units used by the transaction (gas price in tx)
     Amount fee = 4;                 // transaction fee charged for the transaction
     uint64 layer_number = 5;        // The layer in which the STF processed this transaction
+    AccountId app_address = 6;      // Deployed app address or code template address !!!
```
lrettig, line 78:
```suggestion
    AccountId app_address = 6;      // Deployed app address or code template address
```

proto/spacemesh/types.proto
```diff
@@ -77,25 +71,45 @@ message TransactionReceipt {
...
+    repeated TransactionReceipt receipts = 2;
+}
+
+// TransactionState is the "journey" of a tx from newly-submitted, to being
+// rejected from or accepted into the mempool, to being included in the mesh (in a
+// block), to (finally) being processed by the STF. It tells us nothing about the
+// result of the processing.
+message TransactionState {
+    enum TransactionStateType {
+        UNKNOWN = 0;                // Default state
```
lrettig, line 92:
```suggestion
        UNDEFINED = 0;              // Default state
```

proto/spacemesh/types.proto
```diff
@@ -109,59 +123,85 @@ message Transaction {
...
 }
 
 message Layer {
-    uint64 number = 1;              // layer number - not hash - layer content may change
     enum LayerStatus {
-        UNDEFINED = 0;
-        UNKNOWN = 1;
-        APPROVED = 2;
-        CONFIRMED = 3;
+        UNKNOWN = 0;
```
lrettig, line 154:
```suggestion
        UNDEFINED = 0;
```

proto/spacemesh/types.proto
```diff
@@ -109,59 +124,85 @@ message Transaction {
...
+
+// Indicates a change in node status
+message NodeStatusEvent {
+    enum NodeStatusEvent {
+      UNDEFINED = 0;
+      SYNCED = 1;                // node is synced
+      SYNCING = 2;               // node is syncing (but not synced)
+      NOT_SYNCING = 3;           // sync did not start
+      SYNC_FAILED = 4;           // node sync fail event. Node will never be syncing or synced again in this session.
+      NEIGHBOR_CONNECTED = 5;    // Connected to a neighbor
```
lrettig, line 183: Is this useful by itself? E.g., without the node id of the new neighbor?

proto/spacemesh/smesher_types.proto
```diff
@@ -0,0 +1,51 @@
...
+    repeated CommitmentProcessor processors = 1;
+}
+
+// Param passed to methods to indicate a request to delete commitment file
+message DeleteCommitment {
+    bool delete_commitment = 1;
+}
+
+// Proof of space commitment data
+message CommitmentData {
```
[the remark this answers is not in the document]
avive: yes, so the idea is at CommitmentUserData is the input data that a user provides when requesting to setup / change a commitment and CommitmentData is an actual commitment data - which has different fields. Maybe the shared fields can be encapsulated in a type and shared between these 2 data structures?

proto/spacemesh/global_state.proto
```diff
@@ -3,17 +3,48 @@ package spacemesh;
...
+    // Rewards awarded to a smesher
+    rpc SmesherRewards(SmesherIdWithOffset) returns (Rewards);
+
+    ////////// Streams
+
+    // Rewards awarded to a specific account
+    rpc AccountRewardStream(AccountId) returns (stream Reward);
+
+    // Rewards awarded to a smesher id
+    rpc SmesherRewardStream(SmesherId) returns (stream Reward);
```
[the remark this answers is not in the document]
avive: no - this will be very wasteful. Wallets are only interested in a small subset of rewards, not that 50 or so rewards every 5 minutes that the protocol computes! So, this is needed to get rewards that the wallet knows it needs to display in the future without polling. I'm not sure this can be done otherwise. We need 2 methods because users may change the coinbase account for rewards and want to know of all rewards to their miner regardless of these changes...

proto/spacemesh/global_state.proto
```diff
@@ -3,17 +3,48 @@ package spacemesh;
...
+
+    // Returns a stream of Account changes. e.g. balance and counter
+    // (nonce) changes
+    rpc AccountStream(AccountId) returns (stream Account);
```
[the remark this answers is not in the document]
avive: This will create a stream that outputs a data item on each account balance change in the system and not just for accounts that the wallet is interested at. It will be very wasteful and the wallet will need to process each incoming message and check if it is for one of the addresses it is interested at - instead - the pattern is that it subscribes just on the accounts it is interested in knowing about and only get applicable messages that it cares about when they happen over the stream - much more efficient, less messages and less processing...

proto/spacemesh/global_state.proto
```diff
@@ -3,17 +3,48 @@ package spacemesh;
...
 import "spacemesh/types.proto";
 
-// Readonly global state data.
+// Readonly global state data - current and historical.
 // Global state data is data which is not explicitly stored in the mesh.
 // Global state is modified only by the state transition function.
 service GlobalStateService {
-    //// Streaming global state data
+    // Latest computed global state - layer and its root hash
+    rpc GlobalState(google.protobuf.Empty) returns (GlobalState);
```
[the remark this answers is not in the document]
avive: yes, so the idea is that wallets display the last known global state and its layer for display to users. As we don't have consensus on this state, we want at least for users to see it so they can compare with each other and provide it to us in bug reports until we have full telemetry for it from nodes... We already display it in the wallet settings today but w/o a layer id it is useless information - just a random hash...

proto/spacemesh/mesh.proto
```diff
@@ -20,9 +21,21 @@ service MeshService {
...
+    rpc AccountTransactions(AccountIdWithOffset) returns (Transactions);
+
     ////////// Streams
 
-    // Layer with blocks and transactions.
-    // Sent each time layer data changes. Designed for heavy-duty clients.
+    // A stream of new txs (or update to a tx data) where the account is sender
+    // or receiver. These should include activation transations (ATXs) and all
+    // transactions in the mesh
+    rpc AccountTransactionStream(AccountId) returns (stream Transaction);
```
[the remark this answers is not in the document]
avive: As I commented above - this will be extremely wasteful in terms of number of messages that are sent and needs to be processed by clients. Wallets are only interested in a very small subset of accounts and not in every transaction in the network. Without this - clients will have compare all the accounts they care about with each transaction in the network - constant processing instead of just getting the data they need and acting upon it...

proto/spacemesh/types.proto
avive: We said we want it to support pagination. I don't know how else to do pagination besides adding offset and optionally number of requested results to client requests. It will need to be evaluated against the servers notion of a set of matching items.

proto/spacemesh/types.proto
avive: probably not very useful. I was trying to get away from adding a separate network status events streams. So the idea is that when such a status event is received, node client can query the node state for number of neighbors currently connected to figure out the net status. This is just basically a hint that p2p connectivity status changed. Maybe not the  best way to do it. Maybe it is better to just have NET_CHANGED event and clients may decide to query the node for additional net info when they get such an event from a node. Again - instead of having to constantly poll...

proto/spacemesh/smesher_types.proto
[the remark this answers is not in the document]
lrettig: sounds good to me. let me know if you want to make this change or if I should make it. also, can we add this explanation to the comments to make it clearer?

proto/spacemesh/global_state.proto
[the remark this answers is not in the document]
lrettig: The requirements for the wallet makes sense, but I still think it may be easier to write this filtering logic on the frontend than inside go-spacemesh. We might want to put this question before the other devs, in a design review.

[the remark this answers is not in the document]
lrettig: Let's take this to #26

proto/spacemesh/global_state.proto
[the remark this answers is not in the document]
lrettig: Let's take this to #26

proto/spacemesh/global_state.proto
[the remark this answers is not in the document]
lrettig: Okay, so there's no need to get global state as of a certain previous layer? Got it. Reopen if I got that wrong.

proto/spacemesh/mesh.proto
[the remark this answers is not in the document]
lrettig: #26

proto/spacemesh/types.proto
lrettig: Actually, I take this back. The state is in the API call, and doesn't necessarily need to be stored on the backend. The backend can just reload everything and respond from the offset.

proto/spacemesh/types.proto
lrettig: Okay

proto/spacemesh/smesher_types.proto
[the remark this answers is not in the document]
avive: You can do it.

proto/spacemesh/global_state.proto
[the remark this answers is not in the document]
avive: ok

proto/spacemesh/smesher_types.proto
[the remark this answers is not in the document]
lrettig: Done

lrettig:
`buf` reports a bunch of breaking changes between the existing v0.1 API (`master` branch) and these changes. What do we want to do about these? Here's a full report:

```
> buf check breaking --against-input image.bin
proto/spacemesh/global_state.proto:40:23:RPC "AccountStream" on service "GlobalStateService" changed request type from ".google.protobuf.Empty" to ".spacemesh.AccountId".
proto/spacemesh/types.proto:92:5:Previously present enum value "6" on enum "TransactionStateType" was deleted without reserving the name "PROCESSED".
proto/spacemesh/types.proto:92:5:Previously present enum value "6" on enum "TransactionStateType" was deleted without reserving the number "6".
proto/spacemesh/types.proto:94:20:Enum value "1" on enum "TransactionStateType" changed name from "UNKNOWN" to "REJECTED".
proto/spacemesh/types.proto:95:30:Enum value "2" on enum "TransactionStateType" changed name from "REJECTED" to "INSUFFICIENT_FUNDS".
proto/spacemesh/types.proto:96:23:Enum value "3" on enum "TransactionStateType" changed name from "INSUFFICIENT_FUNDS" to "CONFLICTING".
proto/spacemesh/types.proto:97:19:Enum value "4" on enum "TransactionStateType" changed name from "CONFLICTING" to "PENDING".
proto/spacemesh/types.proto:98:21:Enum value "5" on enum "TransactionStateType" changed name from "PENDING" to "PROCESSED".
proto/spacemesh/types.proto:111:21:Enum value "4" on enum "TransactionType" changed name from "DEPLOY" to "SPAWN_APP".
proto/spacemesh/types.proto:154:5:Previously present enum value "3" on enum "LayerStatus" was deleted without reserving the name "CONFIRMED".
proto/spacemesh/types.proto:154:5:Previously present enum value "3" on enum "LayerStatus" was deleted without reserving the number "3".
proto/spacemesh/types.proto:156:20:Enum value "1" on enum "LayerStatus" changed name from "UNKNOWN" to "APPROVED".
proto/spacemesh/types.proto:157:21:Enum value "2" on enum "LayerStatus" changed name from "APPROVED" to "CONFIRMED".
proto/spacemesh/types.proto:168:1:Previously present field "6" with name "current_layer" on message "NodeStatus" was deleted without reserving the name "current_layer".
proto/spacemesh/types.proto:168:1:Previously present field "7" with name "verified_layer" on message "NodeStatus" was deleted without reserving the name "verified_layer".
proto/spacemesh/types.proto:168:1:Previously present field "6" with name "current_layer" on message "NodeStatus" was deleted without reserving the number "6".
proto/spacemesh/types.proto:168:1:Previously present field "7" with name "verified_layer" on message "NodeStatus" was deleted without reserving the number "7".
proto/spacemesh/types.proto:169:5:Field "1" with name "connected_peers" on message "NodeStatus" changed option "json_name" from "knownPeers" to "connectedPeers".
proto/spacemesh/types.proto:169:12:Field "1" on message "NodeStatus" changed name from "known_peers" to "connected_peers".
proto/spacemesh/types.proto:170:5:Field "2" with name "is_synced" on message "NodeStatus" changed option "json_name" from "minPeers" to "isSynced".
proto/spacemesh/types.proto:170:5:Field "2" on message "NodeStatus" changed type from "uint64" to "bool".
proto/spacemesh/types.proto:170:12:Field "2" on message "NodeStatus" changed name from "min_peers" to "is_synced".
proto/spacemesh/types.proto:171:5:Field "3" with name "synced_layer" on message "NodeStatus" changed option "json_name" from "maxPeers" to "syncedLayer".
proto/spacemesh/types.proto:171:12:Field "3" on message "NodeStatus" changed name from "max_peers" to "synced_layer".
proto/spacemesh/types.proto:172:5:Field "4" with name "top_layer" on message "NodeStatus" changed option "json_name" from "isSynced" to "topLayer".
proto/spacemesh/types.proto:172:5:Field "4" on message "NodeStatus" changed type from "bool" to "uint64".
proto/spacemesh/types.proto:172:12:Field "4" on message "NodeStatus" changed name from "is_synced" to "top_layer".
proto/spacemesh/types.proto:173:5:Field "5" with name "verified_layer" on message "NodeStatus" changed option "json_name" from "syncedLayer" to "verifiedLayer".
proto/spacemesh/types.proto:173:12:Field "5" on message "NodeStatus" changed name from "synced_layer" to "verified_layer".
proto/spacemesh/types.proto:202:21:Enum value "3" on enum "NodeErrorType" changed name from "SIGNAL_SHUT_DOWN" to "PANIC_P2P".
proto/spacemesh/types.proto:206:5:Field "1" with name "error_type" on message "NodeError" changed option "json_name" from "type" to "errorType".
proto/spacemesh/types.proto:206:19:Field "1" on message "NodeError" changed name from "type" to "error_type".
```

avive:
I think we should not worry about 0.1 compatibility because everything is in a flux now and not yet even pass design review. I think it is highly likely that even this will still change in many ways based on design reviews and found issues...

proto/spacemesh/smesher.proto
```diff
@@ -0,0 +1,54 @@
...
+    // Returns post commitment data from the node
+    rpc CommitmentData(google.protobuf.Empty) returns (CommitmentData);
+
+    // Returns a list of available commitment init processors
+    // (including CPU as fallback)
+    rpc CommitmentProcessors(google.protobuf.Empty) returns (CommitmentProcessors);
+
+    // Starts (or continues) a commitment init phase. Supports appending to file
+    // from a previous incomplete init phase.
+    rpc CreateCommitment(CommitmentUserData) returns (BoolResult);
```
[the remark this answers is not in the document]
avive: yes

proto/spacemesh/smesher.proto
```diff
@@ -0,0 +1,54 @@
...
+
+    // Stop an ongoing commitment init phase and optionally attempt to delete
+    // commitment file(s)
+    rpc StopCommitmentCreation(DeleteCommitment) returns (BoolResult);
```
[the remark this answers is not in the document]
avive: Good comment. See - comment above. I think we should change to return a typed message with a boolean result value and extend it to return errors.

PR #32, opened 2020-05-31
Refactor Transaction to proper types

### LINKED ISSUE #27: ATX model missing fields
From [Slack standup](https://spacemesh.slack.com/archives/CREMHKGT0/p1590669101025700) today:

Alina Tustanovska: 4. [ATXs]: Wold you like us to add ActiveSetSize param to Transaction, or even better to define ATX as a separate type and add the ATXs field to the Block?

Alina Tustanovska:
type ActivationTxHeader struct {
 NIPSTChallenge
 id      *ATXID // non-exported cache of the ATXID
 Coinbase   Address
 ActiveSetSize uint32
}
type NIPSTChallenge struct {
 NodeID        NodeID
 Sequence       uint64
 PrevATXID      ATXID
 PubLayerID      LayerID
 StartTick      uint64
 EndTick       uint64
 PositioningATX    ATXID
 CommitmentMerkleRoot []byte
}

i.e., this is part of the ATX struct in go-spacemesh but isn't currently captured by the API.

CC @primaveralina

### LINKED ISSUE #31: Refactor transaction data item
We should use a typed message for each tx type. e.g. simple coin / smart contract and atx and use the protobufs `oneof` feature to include 1 tx data in a transaction object. Right now the ATX fields are a hack in the tx message. Also - we should use optional for optional ATX fields if we keep the current Transaction message the way it is.

avive:
Much better not to mix tx types data in one object - now using oneof.
If we unify the simple coin tx with svm tx the we still have 2 types - atx and svm tx - so this will be useful. Also added signature to transaction as this is part of the tx immutable mesh data.

Supersedes #30
Closes #27
Closes #31


lrettig:
This closes #31 right?

avive:
yep

proto/spacemesh/types.proto
```diff
@@ -103,28 +103,54 @@ message TransactionState {
...
+
+// A simple signature data
+message Signature {
+    enum Scheme {
+        ED25519 = 0;            // Standard Ed25519 scheme
+        ED25519_PLUS_PLUS = 1;  // Spacemesh-modified ED25519 scheme, a.k.a. ED25519++
+    }
+    Scheme scheme = 1;    // The signature's scheme
+    bytes signature = 2;  // The signature itself
+    bytes public_key = 3; // Included in schemes which require signer to provide a public key
```
lrettig, line 137: How do we link this to AccountId/address? Do we want to do that here, or just leave them distinct?

proto/spacemesh/types.proto
```diff
@@ -103,28 +103,54 @@ message TransactionState {
...
-    TransactionFee fee = 4;
-    uint64 timestamp = 5;  // shouldn't this be part of the event envelope?
-    AccountId receiver = 6; // depending on tx type
-    Amount amount = 7; // amount of coin transfered in this tx by sender
-    uint64 counter = 8;    // tx counter aka nonce
-    bytes data = 9; // binary payload - used for app, deploy, atx and spwan transactions
-    SmesherId smesher_id = 10;    // used in atx only
-    TransactionId prev_atx = 11;  // previous ATX. used in atx.
+    AccountId sender = 5;
+    TransactionFee fee = 6;
```
lrettig: Shouldn't this be gas price, and max gas?

proto/spacemesh/types.proto
```diff
@@ -103,28 +103,54 @@ message TransactionState {
...
+    uint64 timestamp = 7;  // shouldn't this be part of the event envelope?
```
lrettig: Is this the timestamp when the tx is created, broadcast, processed, or something else? I'm not sure we need this here. Tx should be immutable and the time when it's processed/included can't be known when it's created.

lrettig:
You know what I'm going to say about the enums... made a few small points above, otherwise looks good. Clever use of `oneof`!

proto/spacemesh/types.proto
avive: I don't understand the question. what is this?

lrettig: Public key is not the same thing as AccountId - how do we cross-reference the two?

avive: I think that the public key is only interesting in the context of a signature that includes it. e.g. ED25519 as this data will be on the mesh. It is a good question regarding the need to include account id in coin transactions that use this scheme as it can be extracted from the public key...

avive: Maybe we create it from the public key and it doesn't need to be included in the tx itself but when we create a transaction data item we populate it with the account id for convenience

proto/spacemesh/types.proto
avive: It is hard for me to understand this comment due to my changes - is this still applicable with the most recent proposed changes to types?

proto/spacemesh/types.proto
avive: good question. I'm not sure about this as well. Maybe it is don't needed.

avive: I removed it for now.

proto/spacemesh/types.proto
avive: I think we should live then distinct even though the account is the 20 bytes out of the key.

proto/spacemesh/types.proto
avive: I fixed it. TransactionFee should be user provided gas data. I think provided and max are the same thing so you are correct.

proto/spacemesh/types.proto
avive: I and I think they are cross-referenced by the fact the Signature is a member of Transaction always has an AccountId for sender.

avive:
> You know what I'm going to say about the enums... made a few small points above, otherwise looks good. Clever use of `oneof`!

Maybe we should get an arbtrar to achieve consensus?

proto/spacemesh/types.proto
lrettig: I don't think we'd ever want to show the signature, or the public key, inside the downstream, user-facing apps such as wallet/explorer. We'll want AccountId (address) instead. That's why I'm asking. Do we want the backend to convert from public key -> address, and include this explicitly, or do we expect the downstream apps to do the public key conversion/extraction themselves?

avive: there's another field for the account so they don't need to extract. I think it is useful to include the signature in transactions but I don't feel strongly about it. It will be useful to do analysis regarding what signature scheme used and for which transaction. It can be extracted from txs and the mesh by any full node anyhow so this is just convenience.

proto/spacemesh/types.proto
```diff
@@ -66,13 +58,11 @@ message TransactionReceipt {
     TransactionId id = 1;           // the source transaction
     // The results of STF transaction processing
     enum TransactionResult {
```
lrettig, line 60: While I agree with you that this enum represents the output/result of the STF, and therefore must have a well-defined value, I disagree with removing the `UNDEFINED` value here simply because the value of a newly-instantiated data structure should not default to, e.g., `EXECUTED`. That could lead to hard-to-detect bugs. It should default to `UNDEFINED` at instantiation time and need to be subsequently set.

proto/spacemesh/types.proto
lrettig: ok

proto/spacemesh/types.proto
avive: I just can't imagine a case where the value of this enum will be undefined in its container message. It is not a valid value for this enum - when the node creates a receipt this enum must have one of the defined values. Adding another unused value actually can introduce bugs where the code that prepares a receipt forgets to set it. Requiring it to be explicit is a good thing in my opinion.

avive: but if you feel strongly about it then I can revert.

lrettig: > Requiring it to be explicit is a good thing in my opinion

I think we're in agreement here aren't we? My point is exactly this: that it should have to be explicitly set. Its default value should be an error state.

> It is not a valid value for this enum - when the node creates a receipt this enum must have one of the defined values

Exactly, I agree that it's not a valid value - which makes it easier to detect errors! If it ever has this value, then there's a bug in the system somewhere.

lrettig: > Adding another unused value actually can introduce bugs where the code that prepares a receipt forgets to set it

I actually think it makes this case less likely. If a receipt forgets to set it, and it defaults to a _valid_ value, that's much harder to spot than its defaulting to an invalid value

avive: I added it back for now.

PR #43, opened 2020-06-03
Mesh api update

avive:
Based on review comments - added support for stream filtering and for non-streaming queries


avive:
I was bad and based this off the smesher pr - so closing and doing a new one

Issue #40, opened 2020-06-02
Muxing and streams


lrettig:
Rather than one stream per type, we may want to switch to a single master event stream that a client can subscribe to. The client can optionally pass a filter when setting up the stream if they only want to receive a subset of event types. Some advantages:

- this is easier to handle on the client side - rather than needing, e.g., seven logical threads to monitor seven types of events, the client can use a single logical thread (with a case/select statement) and demux/dispatch handling of different event types
- it maintains optionality since, if the current behavior is desired, the client can just create seven different event streams, each filtering on a specific event type

We should look into whether there's an existing specification language/DSL we can use for the filtering/stream definition.

avive:
We have a real api customer on the team - @IlyaVi - we should ask with him what he prefers as a wallet builder and consider his feedback. Also - per @antonlerner feedback - we should look into pub-sub vs. grpc for the api implementation framework.

lrettig:
Looking into pub sub is on my to do list

lrettig:
This work has been done. Results are in https://github.com/spacemeshos/SMIPS/issues/13 and https://github.com/spacemeshos/SMIPS/issues/21. We ended up implementing multiplexed streams in the API (see linked, merged PRs). I looked into pubsub and that conversation is ongoing in the implementation SMIP. We don't have any fancy DSL or anything for filtering now but we can look into that in future if necessary.

@ilyavi @ilans @amitshaul please let us know if you need any changes made to the streams as defined!

Issue #37, opened 2020-06-02
Handling queries of historical data (Streams vs. Queries)


lrettig:
GRPC streams are well-suited to _live_ data, but are less well-suited to retrieving historical data for several reasons:

- backend implementation is harder (stateful)
- if you want, say, just one event that occurred recently in a table of millions of events, you'd have to consume the entire stream to get the recent data
- sometimes pagination just makes more sense since it better aligns with the client use case, e.g., showing actual pages of transactions in an explorer - it's unclear how the client would consume data in pages over a stream, as there are challenges with requesting more data, buffering, etc.

Also: whether a particular endpoint returns data as pages, or as a stream, is a separate question from whether it acts as a logical Stream (with a capital S), i.e., puts all events on the wire as soon as it knows about them (as opposed to a Query, which fetches historical data).

It's likely we'll want Queries for querying historical data, in addition to Streams for receiving live data.

Related: #33

PR #44, opened 2020-06-03
Add queries and stream filters

avive:
Per design review.

Partially addresses #37 
Cf. #40

### DIFF
--- a/proto/spacemesh/mesh.proto
+++ b/proto/spacemesh/mesh.proto
@@ -26,6 +26,9 @@ service MeshService {
     // Number of transactions per second (a network parameter)
     rpc MaxTransactionsPerSecond(google.protobuf.Empty) returns (SimpleInt);
 
+    ////////// Queries
+    // Queries return paginated, historical data.
+
     // Get account data query
     rpc AccountData(AccountMeshDataQueryParams) returns (AccountMeshDataResult);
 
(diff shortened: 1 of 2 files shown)


proto/spacemesh/mesh.proto
```diff
@@ -24,16 +26,21 @@ service MeshService {
     // Number of transactions per second (a network parameter)
     rpc MaxTransactionsPerSecond(google.protobuf.Empty) returns (SimpleInt);
 
-    ////////// Streams
+    // Get account data query
+    rpc AccountData(AccountMeshDataQueryParams) returns (AccountMeshDataResult);
```
lrettig, line 33:
```suggestion
    rpc AccountData(AccountMeshDataQueryParams) returns (AccountMeshDataQueryResult);
```

proto/spacemesh/mesh_types.proto
```diff
@@ -0,0 +1,44 @@
...
+}
+
+message AccountMeshData {
+    oneof data_item {
+        Transaction transction = 1;
+        Activation activation = 2;
+    }
+}
+
+message AccountMeshDataResult {
```
lrettig, line 32:
```suggestion
message AccountMeshDataQueryResult {
```

proto/spacemesh/mesh.proto
avive: I think it is better and cleaner to use a bit field for filters and not a list of filters. It shouldn't be a problem to create them in any modern client / language. Yes, global state changes should be in its own prs to keep prs small and about less things...

Issue #36, opened 2020-06-01
Refactor Transaction Receipt/Result/State


lrettig:
Per today's design review call, Noam and Iddo both questioned the current design of `TransactionResult`. Maybe what we really want to know is: 1. was the transaction "effective" or not (Iddo's term, i.e., syntactically valid), 2. did the transaction succeed or fail?, and 3. was a fee charged for it or not? These things are not immediately visible from the current design. Do we want to rethink the design?

avive:
I'm a bit confused about this issue. 

A transaction is effective / syntactically valid if it made its way to the mesh and invalid if it was rejected before that as there are validity checks when it is submitted by a client to a node for broadcasting. See TransactionState.REJECTED.

I think that as soon as it is TransactionState.PENDING on its journey to the mesh, it means it passed syntactic validation and is in the pool. 

The `TransactionState` enum is design to capture all possible rejections which can be considered types of failures before it gets to STF execution. What's may be confusing to reviewers here is that there are pre-stf validation checks (designed by Noam) so a transaction may be rejected by-stf due to non-syntactic reasons such as counter and balance. However, all of these failures are captured by TransactionState.

@noamnelke @tal-m

Regarding execution failure. The receipt should have all the information regarding the other questions in the issue which are different questions than syntactic validity. So, I don't understand the concern described based on the current design - it is immediately clear exactly what is the transaction status from the receipt, and the TransactionStatus enum should cover all possible cases and questions raised above regarding execution and there's also a field in the receipt for how much gas was consumed. So that is fully addressed.

A receipt is generated even if execution 'fails' because gas is consumed and the tx affected the global state in that way, even though the intent of the tx was not `applied`. So, even a transaction that was no `applied` - e.g funds were not moved according what it specifies, affects the global state and generates a receipt. This is good because the receipt is the only way for clients to know when it was executed (layer # of STF) and why it wasn't applied - and TransactionResult field tells you exactly that.

I think Iddo and Tal argued we should expose the ordered list of input transactions to the STF in each layer via the global state API, and have an index in the tx receipt that indicates the location of the tx in this ordered list so we can show users what txs were executed before or after it in the layer. As the ordered list is per layer, this gives us all the ordered txs across all layers which Tal mentioned.

lrettig:
> A transaction is effective / syntactically valid if it made its way to the mesh and invalid if it was rejected before that as there are validity checks when it is submitted by a client to a node for broadcasting.

[...]

PR #45, opened 2020-06-04
Global state api patch (WIP)

### LINKED ISSUE #37: Handling queries of historical data (Streams vs. Queries)

avive:
@lrettig - please review and let's talk.

Closes #37 
Cf #40

### DIFF
--- a/proto/spacemesh/global_state.proto
+++ b/proto/spacemesh/global_state.proto
@@ -18,7 +18,7 @@ service GlobalStateService {
     // Returns the current state of one or more transactions
     rpc TransactionsState(TransactionsIds) returns (TransactionsState);
 
-    // Account info in the current global state
+    // Account info in the current global state.
     rpc Account(AccountId) returns (Account);
 
     // Query for account related data such as rewards, tx receipts and account info
@@ -51,5 +51,5 @@ service GlobalStateService {
     // a specific app instance...
 
     // New global state computed for a layer by the STF
-    rpc GlobalStateStream(GlobalStateParams) returns (stream GlboalStateStreamResult);
+    rpc GlobalStateStream(GlobalStateEntities) returns (stream GlboalStateEntitiesResponse);
 }
--- a/proto/spacemesh/global_state.proto
+++ b/proto/spacemesh/global_state.proto
@@ -11,10 +11,6 @@ service GlobalStateService {
     // Latest computed global state - layer and its root hash
     rpc GlobalState(google.protobuf.Empty) returns (GlobalState);
 
-    // Returns current transaction state which describes its journey
-    // to the mesh and from the mesh to the global state (if it got processed)
-    rpc TransactionState(TransactionId) returns (TransactionState);
-
     // Returns the current state of one or more transactions
     rpc TransactionsState(TransactionsIds) returns (TransactionsState);
 
--- a/proto/spacemesh/global_state.proto
+++ b/proto/spacemesh/global_state.proto
@@ -9,7 +9,7 @@ import "spacemesh/global_state_types.proto";
 // Global state is modified only by the state transition function.
 service GlobalStateService {
     // Latest computed global state - layer and its root hash
-    rpc GlobalState(google.protobuf.Empty) returns (GlobalState);
+    rpc GlobalStateHash(google.protobuf.Empty) returns (GlobalStateHash);
 
     // Account info in the current global state.
     rpc Account(AccountId) returns (Account);
--- a/proto/spacemesh/global_state.proto
+++ b/proto/spacemesh/global_state.proto
@@ -11,9 +11,6 @@ service GlobalStateService {
     // Latest computed global state - layer and its root hash
     rpc GlobalState(google.protobuf.Empty) returns (GlobalState);
 
-    // Returns the current state of one or more transactions
-    rpc TransactionsState(TransactionsIds) returns (TransactionsState);
-
     // Account info in the current global state.
     rpc Account(AccountId) returns (Account);
 
--- a/proto/spacemesh/global_state.proto
+++ b/proto/spacemesh/global_state.proto
@@ -51,5 +51,5 @@ service GlobalStateService {
     // a specific app instance...
 
     // New global state computed for a layer by the STF
-    rpc GlobalStateStream(GlobalStateEntities) returns (stream GlboalStateEntitiesResponse);
+    rpc GlobalStateStream(GlobalStateStreamParams) returns (stream GlboalStateStreamResponse);
 }
--- a/proto/spacemesh/global_state.proto
+++ b/proto/spacemesh/global_state.proto
@@ -2,6 +2,7 @@ syntax = "proto3";
 package spacemesh;
 import "google/protobuf/empty.proto";
 import "spacemesh/types.proto";
+import "spacemesh/global_state_types.proto";
 
 // Readonly global state data - current and historical.
 // Global state data is data which is not explicitly stored in the mesh.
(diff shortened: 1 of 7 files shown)


proto/spacemesh/global_state_types.proto
```diff
@@ -0,0 +1,71 @@
...
+// Data items related to an account
+enum AccountDataFlag {
+    NONE = 0;
+    TRANSACTION_RECIEPT = 1; // tx receipt for a tx to or from an account
+    REWARD = 2; // reward awarded to an account
+}
+
+message AccountDataParams {
+    AccountId account_id = 1;
+    uint32 account_data_bit_field = 2; // A bit field of AccountDataFlag
```
lrettig, line 14: This was called `flags` in mesh_types, can we harmonize the naming?

https://github.com/spacemeshos/api/blob/321c19757e5381f56c306f1eb99660ba72f819ae/proto/spacemesh/mesh_types.proto#L14

proto/spacemesh/global_state_types.proto
```diff
@@ -0,0 +1,71 @@
...
+}
+
+message AccountDataQueryParams {
```
lrettig, line 17: No `min_layer`?

proto/spacemesh/global_state.proto
```diff
@@ -10,45 +11,45 @@ service GlobalStateService {
...
-    // Get the current transaction state. Note that tx state is not part of
-    // global state. It is metadata about a transaction on its journey to the
-    // mesh and from the mesh to the global state (or to null)
+    // Returns current transaction state which describes its journey
+    // to the mesh and from the mesh to the global state (if it got processed)
     rpc TransactionState(TransactionId) returns (TransactionState);
 
-    // Account info in the current global state
+    // Returns the current state of one or more transactions
+    rpc TransactionsState(TransactionsIds) returns (TransactionsState);
```
lrettig, line 19: What's the advantage of having this, over just allowing repeated calls to `TransactionState`?

proto/spacemesh/global_state_types.proto
```diff
@@ -0,0 +1,71 @@
...
+    TRANSACTION_STATE = 2;
+    TRANSACTION_RECIEPT = 4;
+    GLOBAL_STATE = 8;
+}
+
+message GlobalStateEntities {
+    uint32 flags = 1; // bitmap of GlobalStateEntityFlags
+}
+
+message GlobalStateEntity {
```
lrettig, line 60: I don't think `Entity` is the right word here - maybe `Entry`? Or `Object` or `Datum`

proto/spacemesh/mesh_types.proto
```diff
@@ -3,7 +3,7 @@ package spacemesh;
 
 import "spacemesh/types.proto";
 
-enum AccountMeshDataFlags {
```
lrettig, line 6: I think this should be plural... one bitfield contains many flags

proto/spacemesh/global_state_types.proto
```diff
@@ -0,0 +1,71 @@
...
+    uint32 max_results = 2; // max numbers of results client would like to get
+    uint32 offset = 3; // return results from offset
+}
+
+// An account related data item
+message AccountData {
+    oneof item {
+        Reward reward = 1;
+        TransactionReceipt receipt = 2;
+        Account account = 3;
```
lrettig, line 28: Why not separate out the various event types encapsulated in an update to Account? E.g., NewNonce, NewBalance, NewCode, etc.

Also, it sort of bothers me that the event types contained in this oneof/enum actually overlap: e.g., a `Reward` event will correspond to a `NewBalance` event. Is there some more intuitive way to bundle these?

proto/spacemesh/global_state_types.proto
```diff
@@ -0,0 +1,71 @@
...
+    }
+}
+
+message AccountDataQueryResults {
+    repeated AccountData account_item = 1;
+    uint32 total_results = 2;
+}
+
+message SmesherDataQueryParams {
```
lrettig, line 37: No `min_layer`?

proto/spacemesh/global_state_types.proto
```diff
@@ -0,0 +1,71 @@
...
+}
+
+message GlobalStateEntities {
+    uint32 flags = 1; // bitmap of GlobalStateEntityFlags
+}
+
+message GlobalStateEntity {
+    oneof data {
+        Reward reward = 1;
+        TransactionState transaction_state = 2;
```
lrettig, line 63: Strictly speaking I don't think `TransactionState` belongs in the global state. Isn't it kinda orthogonal to global state? Cf. #36

proto/spacemesh/smesher_types.proto
```diff
@@ -1,6 +1,6 @@
 syntax = "proto3";
 package spacemesh;
-
+import "spacemesh/types.proto";
```
lrettig, line 3: Doesn't seem required, is it?

proto/spacemesh/global_state_types.proto
avive: I think we should use descriptive var names so I'll change flags to something better.

proto/spacemesh/global_state_types.proto
avive: Good question. See my comments about the queries in GlobalStateService. I didn't add min layer to some queries because this assumed that all of these data items will needed to be indexed by account id and and list support min layer ordering (or a layer index). I think we should have them w/o min layer for now until you find out how expensive it is to support the more robust version with min_layer for them...

proto/spacemesh/global_state.proto
avive: avoiding having to issue multiple calls to TransactionState() when client has more than one tx id it wants results for. And the other version when state of only 1 transaction is expected. So users won't have to use TransactionsState() in this case... it is not huge deal - I think it is nice and anticipates the calling patterns.

proto/spacemesh/global_state_types.proto
avive: I ran out of good names today. These are global state entities. I think entity describes these better than the other names you propose. Maybe `Item` or `DataItem` is better?

proto/spacemesh/mesh_types.proto
avive: I had a look at the naming pattern of enums in the proto3 google docs examples and they all seem to be singular - agree we should use 1 style and be consistent. I think that right now all enums are singular, or at least should be.

proto/spacemesh/global_state_types.proto
avive: Good idea but it requires more work on the server-side - fire an event every time a field changes instead of firing one event after these change. It is going to be very common that counter and balance change - every time there's a tx from that account. I think it is okay by now and moves the responsibility for clients to update their data with the new account data. I don't think strongly about this but I think it is good enough for the first version.

proto/spacemesh/global_state_types.proto
avive: See my comment above regarding the comments in global_state.proto about min_layer support for account queries.

proto/spacemesh/global_state_types.proto
avive: You are correct here. Looking at TransactionStateType I see it has one state that is STF related (PROCESSED) that we should remove and if we do that then it is only a mesh related tx state as it should be. I will fix it and remove it.

proto/spacemesh/smesher_types.proto
avive: hm.. It is removed in my branch - maybe I need to push a bit more changes.

proto/spacemesh/global_state_types.proto
avive: I've changed all `flag` fields to more descriptive names.

proto/spacemesh/global_state_types.proto
avive: Changed to DataItem and cleaned up the types a bit.

proto/spacemesh/global_state_types.proto
lrettig: Okay, I'm happy to keep these simpler for now, until we know how hard implementation will be as you suggest.

proto/spacemesh/global_state_types.proto
lrettig: I'm still confused by this message. I'd expect to see a stream of all events that originate from or touch an account: updates to nonce and balance, rewards, transaction state changes for sent transactions, and maybe a list of transactions for incoming transactions. I don't understand what the 'Account' message is doing in this case. It seems to be a convenience wrapper around nonce + balance changes, but I'd rather include those as first class data items here. Also incoming transactions seems to be missing.

proto/spacemesh/global_state_types.proto
```diff
@@ -0,0 +1,71 @@
...
+    uint32 max_results = 2; // max numbers of results client would like to get
+    uint32 offset = 3; // return results from offset
+}
+
+message SmesherDataQueryResults {
+    uint32 total_results = 1;
+    repeated Reward rewards = 2;
+}
+
+enum GlobalStateDataItemFlag {
```
lrettig, line 48: Why not include Account here?

proto/spacemesh/global_state_types.proto
lrettig: But we have the receipt. If there is no receipt for a tx, then it was never executed by the STF. If there is, then we know its state.

I'm still not totally happy with TransactionState - I think it's very confusing the way it mixes state from a local node (REJECTED, CONFLICTING, etc.), with state from the mesh (PENDING), with post-STF global state (PROCESSED). Maybe let's discuss and work this out in #36?

proto/spacemesh/global_state.proto
lrettig: I see your point but I suppose I prefer keeping the API as simple as possible, and also symmetry across different data types - e.g, why allow a query for multiple transactions but not for multiple accounts?

If we have `TransactionsState` then why do we need `TransactionState` at the same time? Can't you just call `TransactionsState` with a single txid?

avive: ok, I removed the help connivance method.

proto/spacemesh/global_state_types.proto
avive: There is no way around it - because the journey to the mesh passes at these states. It correctly described these possible states. I've removed PROCESSED. So now, it just describes possible states from submission by user until it gets on the mesh. I don't think we want to refactor it into sub-states

avive: It should be in the mesh service and not in the global state service - I've moved it there.

avive: Also added `rpc TransactionsStateStream(TransactionsIds) returns (stream TransactionsState);` which is very useful for wallets to have.

proto/spacemesh/global_state_types.proto
avive: you are correct. it is an error. fixed.

proto/spacemesh/global_state_types.proto
avive: Are you sure we want wallets to get 2 separate events when counter and balance change? it will be very common that these change in the same time when there was an outgoing tx processed.

Regarding transactions results changes you comment about - this is provided the tx receipt event type and it includes incoming coin transfer txs and outgoing transactions.

avive: Also - reward and new balance are not the same - new balance will just show a new balance while reward will show the actual reward and tx fee assigned by an stf in a layer.

lrettig: Maybe my discomfort with this the way it stands here is that it mixes two concerns: _cause_ and _effect._ The _cause_ of a change in the account state is the tx or the reward. The _effect_ is the updated balance/nonce. I can think of two ways to improve this:

1. Bundle cause and effect together into a single 'event' type that always includes both, e.g.,

```
message AccountUpdatedEvent {
    oneof item {
        Reward reward = 1;
        TransactionReceipt receipt = 2;
    }
    Account updated_account_data = 3;
}
```

2. Or else, split this into two different streams and event types, e.g.,

```
message AccountEvent {
    oneof item {
        Reward reward = 1;
        TransactionReceipt receipt = 2;
    }
}
message AccountUpdatedEvent {
    Account updated_account_data = 1;
}
```

Does this make sense?

avive: not really. The way we use this pattern now is that the client specifies what types of events it wants and he can specify he only cares about rewards and/or tx receipts but not in account updates. I really don't see a reason to change it. Rewards and receipts are side-effects and account data change is a change in the account db. I think that the proposed change just over-complicate the api where it is not really needed. If you specify a bitmap of Reward || Account then you should get back both. If you only specify you want to see rewards then you just specify Rewards in the bitmap. In both cases we should give the caller what he asked for.

avive: So the current design is simpler and gives more flexibility to callers.

avive: A similar pattern is now in GlobalStateStream - we don't want to force clients to get account data when they specified they are not interested in this data in the stream.

lrettig: We discussed this on a call just now. I'm still a little uncomfortable with this mix of concerns, but I'm okay with it in the interest of keeping the API simpler, of multiplexing the streams (#40), and as long as it's clearly documented. I documented this more clearly in the comments.

PR #49, opened 2020-06-08
Move TransactionState to NodeTxService

lrettig:
It makes more sense here because it crosses from node to mesh to STF.

Note that this also adds a `PROCESSED` (processed by STF) state back to `TransactionState`.


lrettig:
> It all looks good but looking at this - maybe we should use what your proposed I think today and just call this TransactionService and lose the node part from the name?

Agree, will make this change

PR #57, opened 2020-06-12
Add local golang build

### LINKED ISSUE #55: Add local builds
Rather than requiring that other projects, including go-spacemesh and the explorer backend, include this GitHub repository as a submodule and build the protobuf files as part of their build process, it seems nicer to me if we can add builds to this repository so that they can instead rely upon either 1. those builds, or better yet 2. language-specific modules that are built here. That seems better from the perspective of separation of concerns: those projects should not need to know anything about protobuf, if they're not modifying this API.

I'm investigating how to do this. Here are some options:

- just add a Makefile rule for builds
- trigger a build automatically using a git hook
- require that all PRs, or changes to hit master, include an up-to-date build
- make builds happen automatically as part of CI with every PR/commit
- automatic pre-release when changes land on master (https://github.com/marketplace/actions/automatic-releases)

CC @avive 

Task list:

- [x] golang build command (#57)
- [x] grpc-gateway build command (#57)
- [x] test if this works as a go module (https://github.com/spacemeshos/go-spacemesh/pull/2036)
- [x] make a decision about package naming
- [x] automate the build/prerelease (#62)

lrettig:
Requires #54 (merge that first)
Closes #55


lrettig:
@noamnelke took a look at this and is okay with including the build here, like this. I'm going to go ahead and merge this. We can add other languages as necessary (e.g., swagger, for documentation purposes, or JS, for smapp). Will also start versioning/releasing the builds, and will add a CI check to make sure that commits against master include builds.

noamnelke:
I see we're still using `/v1/` in the path - this is what my comment was about, I don't think we should do that...

PR #71, opened 2020-06-16
Improve go versioning using go.mod

lrettig:
Remove ugly hack that creates dummy files. Add go.mod and go.sum to
build files.


lrettig:
Thanks. For the record, I did test this and I've been using it successfully in go-spacemesh. It appears to have an impact upon how go discovers and names the module and upon where it lives in $GOROOT, but running `go mod tidy` fixed this up.

PR #79, opened 2020-06-24
Submit binary transaction

### LINKED ISSUE #78: Submit binary transaction endpoint
We don't currently have one of these. The `TxService.SubmitTransaction` endpoint accepts a `Transaction` type which is structured and does not contain binary data (aside from the signature):

https://github.com/spacemeshos/api/blob/279f08c40938d979697559aa88bec16c8aab6ff8/proto/spacemesh/v1/types.proto#L105-L116

Per https://github.com/spacemeshos/SMIPS/issues/23 it looks like we need an endpoint to receive signed, binary tx data. Note that the existing, old API does contain an endpoint that does this (`SubmitTransaction`).

CC @avive

avive:
Closes #78


lrettig:
This replaces the existing endpoint, rather than adding a new one. Maybe that's okay. Are we totally certain that we don't want an endpoint that accepts a structured tx, rather than a serialized tx? I suppose, given the need to sign a tx, the client must always have a serialized copy of it?

It strikes me that Ethereum accepts both signed and unsigned transactions, but that's because geth itself has built-in signing capabilities - are we certain that no Spacemesh node will ever do this?

avive:
I think it is okay. 100% sure end point needs to accept binary transactions. Not sure how submitting non-binary Transaction object will work because clients need to generate the tx binary encoded format in order to sign. and the signature is on the binary encoded transaction data... I think that in geth this feature is due to geth's ability to manage private user keys and sign txs on behalf of the users due to historical reasons (it was kind of a wallet) but in spacemesh the full node has no user private keys by design and they are left to be managed by clients such as smapp and cliwallet which I think is a good thing.

avive:
I don't think there's a way to make this change w/o breaking the ci rules as currently defined.

avive:
I've added the built type. Still failing. I think you should override and merge manually.

lrettig:
> Not sure how submitting non-binary Transaction object will work because clients need to generate the tx binary encoded format in order to sign

Okay, I agree. We can always add a new endpoint in the future if need be.

PR #83, opened 2020-06-25
Fix issue 34

### LINKED ISSUE #34: Add state delta and transaction index to TransactionReceipt
Per Tal on today's design review call.

Transaction index is either the index of a tx within a layer (in order of execution, based on output of consensus), or else a global "serial number" of a tx (same thing but across all layers since genesis).

State delta will require some thought. For simple coin tx, it's a list of {accountId: balance change}, as well as the counter (nonce) change for the sender's account. For a smart contract tx, it can additionally include updates to state for one or more smart contracts. For a "create contract" tx, it will also include an update to the newly-created contract address's code.

avive:
Fixes #34


avive:
#34

lrettig:
I'm a bit confused, did you link the right issue here? I don't see a new transaction index nor a state delta

lrettig:
Discussed with Aviv, closing in favor of #95

PR #86, opened 2020-06-26
Add grpc gateway options to all non-stream endpoints

### LINKED ISSUE #85: Finish adding grpc-gateway metadata to all endpoints
Right now only one endpoint has this:

https://github.com/spacemeshos/api/blob/8ce7510395f20fd4cb32f6a2285a97d04966f880/proto/spacemesh/v1/node.proto#L11-L16

lrettig:
Closes #85

Configure the http options using a separate config file rather than
doing it inline in the proto definitions

@avive does this look correct? Do we want to enable this for all endpoints?


lrettig:
Been using this for a few days and it's been working fine, I'll go ahead and merge - @avive let me know if I've missed anything here

PR #91, opened 2020-07-08
Update smesher gpu-post init

avive:
Replace the api-based gpu api with actual enum of available gpus and api availability. 
No way to make this backward compatible. Needed to support more desirable and intuitive behavior and reflect upcoming changes in the gpu post lib api.


avive:
These changes match the updated user's post setup interaction as defined here: https://github.com/spacemeshos/SMIPS/issues/7

avive:
I don't think there's a way to make this change non-breaking. It is fundamental.

proto/spacemesh/v1/smesher.proto
```diff
@@ -41,8 +41,8 @@ service SmesherService {
     // Returns post data status from the node
     rpc PostStatus(google.protobuf.Empty) returns (PostStatusResponse);
 
-    // Returns a bit field of all available setup compute engines
-    rpc AvailableComputeEngines(google.protobuf.Empty) returns (AvailableComputeEnginesResponse);
+    // Returns a list of available gpus for post init
```
moshababo, line 44: perhaps "GPU physical devices and API types"

proto/spacemesh/v1/smesher_types.proto
```diff
@@ -52,20 +52,23 @@ message CoinbaseResponse {
...
-// Bit flags for compute engine selection
-enum ComputeEngineFlag {
-    COMPUTE_ENGINE_FLAG_UNSPECIFIED = 0;
-    COMPUTE_ENGINE_FLAG_CPU = 1; // useful for testing on systems without a cuda or vulkan GPU
-    COMPUTE_ENGINE_FLAG_CUDA = 2;
-    COMPUTE_ENGINE_FLAG_VULKAN = 4;
+message GpuInfo {
+    uint32 id = 1; // 0, 1, 2...
+    string model = 2; // e.g. Nvidia GTX 2700
+    ComputeApi compute_api = 3; // A provided compute api
```
moshababo, line 58: As discussed, there could be multiple API types per physical device, not just one.

proto/spacemesh/v1/smesher_types.proto
```diff
@@ -52,20 +52,23 @@ message CoinbaseResponse {
...
-message ComputeEngineFlags {
-    uint32 compute_engine_flags = 1;
+message GpusResponse {
+    repeated GpuInfo gpu_info = 1;
 }
 
-message AvailableComputeEnginesResponse {
-    ComputeEngineFlags flags = 1;
+// Bit flags for compute engine selection
+enum ComputeApi {
```
moshababo, line 67: Perhaps "ComputeAPIType"

proto/spacemesh/v1/smesher_types.proto
avive: Yes, for each API a device supports you will get an entry with a different id, so for example, on a system with a modern nvidia gpu you will get 2 GpuInfos - 1 for nvidia+cuda and 1 for nvidia+vulkun, each with its own id which is good. When user is setting up post he needs to provide an id of 1 of these entries which specifies both the target gpu and the api. So, I think that the proposed api change fully addresses this issue. I don't think we should return a list of apis supported per gpu - just a tuple (id, gpu_info_string, api) ....

proto/spacemesh/v1/smesher_types.proto
avive: @lrettig - what do you think?

proto/spacemesh/v1/smesher_types.proto
avive: Maybe GPUInfo is confusing and this should just be called ComputeProvider.

avive: I changed GpuInfo to PostComputeProvider - I think this is better now.

moshababo: Got it

proto/spacemesh/v1/smesher_types.proto
```diff
@@ -74,8 +77,8 @@ message PostData {
     string path = 1; // User provided path to create the post data files at
     uint64 data_size = 2; // Requested post data size
     bool   append = 3; // Append to existing files if they exist. Otherwise overwrite.
-    bool   throttle = 4; // Throttle down setup phase computations while user is interactive on ///system
-    ComputeEngineFlags compute_engine_flags = 5; // One or more available compute engines to be used for gpu post
+    bool   throttle = 4; // Throttle down setup phase computations while user is interactive on system
+    uint32 provider_id = 5; // A PostProvider id
```
lrettig, line 81: Why return just an ID here and not a `PostComputeProvider`?

avive: I was thinking about this, it is an option but I didn't add it because the user can easily call PostComputeProviders() and match this id with the response in the case he wants to know what provided was used.

avive: And in addition - this way the full node only needs to persist this id and not the other PostComputeProvider fields.

lrettig: right but what if the list changes and the id no longer maps to the same thing?

avive: it will happen only when you add/remove a gpu and in this case clients can call the method to get the gpus to see what's the current status of the system so I don't think is is a major issue - a bit of an edge case.

PR #94, opened 2020-07-13
Harmonize type names and stream responses

lrettig:
Harmonize the naming of a few types across different services
Update streams so they return scalars not arrays


lrettig:
This has been working fine in the latest API builds, I'm going to go ahead and merge it

PR #95, opened 2020-07-16
Add state delta to tx receipts

### LINKED ISSUE #34: Add state delta and transaction index to TransactionReceipt

lrettig:
Tx index was added already in https://github.com/spacemeshos/api/commit/1b4ca35d24c0eb1aa1ae7980570080442e1bd9f0

This adds the second requirement, state delta

Closes #34

### DIFF
--- a/proto/spacemesh/v1/global_state_types.proto
+++ b/proto/spacemesh/v1/global_state_types.proto
@@ -53,8 +53,24 @@ message TransactionReceipt {
     uint64 gas_used = 3; // gas units used by the transaction
     Amount fee = 4; // transaction fee charged for the transaction (in smidge, gas_price * gas_used)
     uint64 layer_number = 5; // the layer in which the STF processed this transaction
-    uint32 index = 6; // the index of the tx in the ordered list of txs to be executed by stf in the layer.
+    uint32 index = 6; // the index of the tx in the ordered list of txs to be executed by STF in the layer.
     AccountId app_address = 7; // deployed app address or code template address
+    repeated AccountDelta delta = 8; // Side effects from execution of this tx
+}
+
+// All state changes due to this transaction, i.e., side effects in executing
+// the STF on this transaction, including counter, balance, and storage.
+message AccountDelta {
+    AccountId account_id = 1;
+    uint64 new_counter = 2;
+    Amount balance_delta = 3;
+    repeated AccountStorageDelta storage_deltas = 4;
+}
+
+// A change to a single word of account storage
+message AccountStorageDelta {
+    uint64 key = 1;
+    uint64 value = 2;
 }
 
 // All data items that touch an account: receipts for transactions from, or to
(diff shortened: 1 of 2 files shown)


avive:
You should get @YaronWittenstein involved in this and in any API issues such as this which is SVM related - he's been working on svm data in receipts over the last few days...

1. What you propose is a list of decoded key-values. The approach I talked with Yaron and I think what he's building is having packed binary svm data in the receipts where it is the responsibility of clients to decode it (e.g. smapp and explore front-end) using the svm-codec lib. What you proposing will require the node to decode lots of svm data per receipt to create it at request time or at the time it is generated by the node. 

2. We need to talk about the idea to log every variable change in svm. I think that the current thinking was to just include the updated storage root hash in the receipt and so clients can run the computation and compare storage hashes to validate vars change. It is not a bad idea to include (k,v) of every changed value in the receipt.

YaronWittenstein:
I think that `GlobalState` will do that without involving SVM.

This SMIP describes the interface between SVM and the `Global State`:
https://github.com/spacemeshos/SMIPS/issues/20

lrettig:
> just include the updated storage root hash in the receipt and so clients can run the computation and compare storage hashes to validate vars change.

It's not reasonable to expect clients to "run the computation" - they don't have an SVM interpreter nor do they have all the context.

lrettig:
> The `STF` component should be able to ask the `GloblState` for all the `(k, v)` changes and flush to disk at one call all the changes at once.

Great - then it shouldn't be too hard to include this in the tx receipt

YaronWittenstein:
> > just include the updated storage root hash in the receipt and so clients can run the computation and compare storage hashes to validate vars change.
> 
> It's not reasonable to expect clients to "run the computation" - they don't have an SVM interpreter nor do they have all the context.

@lrettig @avive I think I'm not following - I thought that the each Full-Node executes each transaction.
So the Receipt's outcode `App State` is only for validation.

YaronWittenstein:
> > The `STF` component should be able to ask the `GloblState` for all the `(k, v)` changes and flush to disk at one call all the changes at once.
> 
> Great - then it shouldn't be too hard to include this in the tx receipt

As I've written in the previous message here - Why do we need to save all the `(k, v)` changes as part of the Receipt?

lrettig:
> > > just include the updated storage root hash in the receipt and so clients can run the computation and compare storage hashes to validate vars change.
> > 
> > It's not reasonable to expect clients to "run the computation" - they don't have an SVM interpreter nor do they have all the context.
> 
> @lrettig @avive I think I'm not following - I thought that the each Full-Node executes each transaction.
> So the Receipt's outcode `App State` is only for validation.

By client I don't mean full node. I mean something like a web app.

Why include all state changes? Because even a full node can't calculate these for itself without storing all intermediate, historical state.

YaronWittenstein:
But the changes are sitting already in the `Global State` - why replicate that data to the Receipts?

lrettig:
Yes, the global state is updated as a side effect of executing a transaction, but the list of state changes from each tx is not stored by a full node (only maybe an archive node)

fasmat:
Closed due to inactivity

PR #101, opened 2020-07-24
Estimate smeshing rewards endpoint

### LINKED ISSUE #96: Estimate Smeshing Rewards API Method
We'd like to be able to give user an estimate of their rewards in a future time period, say 2 weeks or a month as part of the smesher setup flow. See: https://product.spacemesh.io/#/post_setup . 
To support this feature, we would like to have a new API method which returns this estimate based for example, on the total post size committed to a network in the epoch which just ended.  The value returns should be coin units per wall-clock time unit. e.g. 1 smesh / 30 days. Output can be the estimated rewards amount for the minimum proof of space commitment size, e.g. 100GB (a network param) as clients can use this to compute the expected rewards for multipliers of this data size.
Logically, this method should be added to the Smesher Service.

lrettig:
Closes #96


avive:
I think you want to remove the account id from new methods. The estimated rewards returned should be for the min network storage unit size - a network param. e.g. 100GB and the api is designed to be called by users who don't have an on-mesh account yet. e.g. when setting up post we want to display the estimated rewards per user selected post data size. This is the main use case for this feature. The estimate is not account specific and should return the same value for all clients based on the last epoch's total post data size.

proto/spacemesh/v1/smesher_types.proto
```diff
@@ -129,3 +129,18 @@ message PostStatusResponse {
 message PostDataCreationProgressStreamResponse {
     PostStatus status = 1;
 }
+
+message EstimatedRewardsRequest {
+    AccountId id = 1;
```
avive, line 134: Consider removing

lrettig:
@avive this is ready to go, please take another look

proto/spacemesh/v1/smesher_types.proto
```diff
@@ -129,3 +129,14 @@ message PostStatusResponse {
...
     PostStatus status = 1;
 }
+
+message EstimatedRewardsRequest {}
+
+message EstimatedRewardsResponse {
+    // The amount of the total estimated reward in the next upcoming epoch
+    Amount amount = 1;
+    // Epoch duration
+    SimpleInt duration = 2;
```
avive, line 139: I think this is field is redundant because the epoch duration in time units is part of a network config data that all clients already have.

lrettig: `data_size` should be part of the config, too, but it isn't currently exposed by any API endpoints so I'm going to leave it here for now

PR #102, opened 2020-07-24
Replace error type with error level

### LINKED ISSUE #93: Change NodeErrorType enum to match ordinary debug levels
I'm struggling to map these to the error types in go-spacemesh. I propose that, instead of this relatively arbitrary set of error types, we instead map these to the ordinary set of error levels: debug, info, warn, error, panic, etc.

https://github.com/spacemeshos/api/blob/ff4a9a6b913a23a83cf07a0a7b4b7c16605ce9c0/proto/spacemesh/v1/node_types.proto#L58-L65

@avive where did this list come from? Can we change it?

lrettig:
Closes #93

Levels are based on https://github.com/uber-go/zap/blob/53a387079b46cce51bbb6d360d9c57f86977addb/level.go#L28-L47


avive:
I think there's some confusion here. The intent behind NodeErrorType is for the node to specify a kind of error condition and an not an error so clients can have some info regarding a nature of a specific runtime error and can for example, update users or decide to restart a node they manage. What is the reason behind changing a type that describes possible runtime errors into a generic error level type which doesn't provide details of a specific runtime issue?

lrettig:
@avive see #93 - the logic here is that I'm having trouble mapping all the possible errors the node can produce to a short, relatively arbitrary set of error "types." What's an error "type"? How do we come up with this list of types? How do I classify node errors into this set of types?

lrettig:
@avive this is good to go, could you please review again?

PR #106, opened 2020-07-30
Add Python build and tests

lrettig:
- [x] add python build
- [x] add python build makefile recipe
- [x] add python build makefile recipe to make build
- [x] add python test
- [x] add python test makefile recipe
- [ ] add documentation


avive:
why do we need a python build system? curious...

lrettig:
System tests are all written in Python and need to be migrated. See https://github.com/spacemeshos/go-spacemesh/issues/2185

poszu:
@lrettig, it seems that the python tests are gone. Can this PR be closed?

lrettig:
It's not especially critical but it would still be nice to include a python build here - I think it's likely we'll use python in the future for tooling, testing, simulation, etc. Ideally we'd add typescript and maybe one or two other languages too.

poszu:
> It's not especially critical but it would still be nice to include a python build here - I think it's likely we'll use python in the future for tooling, testing, simulation, etc. Ideally we'd add typescript and maybe one or two other languages too.

Then I'd suggest closing it and reopening it when there is a use case for it.

lrettig:
Okay

PR #107, opened 2020-08-04
Update TransactionReceipt members (breaking change)

avive:
Replace address with svm-data field


lrettig:
To add slightly more context here for posterity, this follows from a conversation we all had a couple of days ago where we discussed the abstraction boundaries around SVM tx/receipt/app data. We decided that we'll include an encoded blob of SVM data in the tx receipt, to be decoded by the client using the svm-codec wasm module (see https://github.com/spacemeshos/svm/tree/master/crates/svm-codec). This is similar to what the client needs to do today with a tx before submitting it using `TxService.SubmitTransaction`.

PR #111, opened 2020-08-10
Split account state into current/projected (breaking change)

### LINKED ISSUE #110: Return both current and projected account data with account endpoints
Motivation: https://github.com/spacemeshos/go-spacemesh/issues/2104

lrettig:
Closes #110 

Also moves account state into global_state_types


lrettig:
@noamnelke this fixes the issue in https://github.com/spacemeshos/go-spacemesh/issues/2104 in the API design, would you mind reviewing? Thanks

PR #103, opened 2020-07-24
Standardize layer number message

lrettig:
Closes #88

PR #120, opened 2020-10-08
Fix layer number type in global state

### LINKED ISSUE #119: layer number type wrong in global state
Looks like I missed these in #103:

https://github.com/spacemeshos/api/blob/9701df00c65a27cf6ea205cd37cc462a78f76b18/proto/spacemesh/v1/global_state_types.proto#L66

https://github.com/spacemeshos/api/blob/9701df00c65a27cf6ea205cd37cc462a78f76b18/proto/spacemesh/v1/global_state_types.proto#L118

lrettig:
Closes #119


lrettig:
CI failure is expected as interface is changing, overriding and merging. This PR fixes a few more lingering references to layer ID type that should've been caught in #103.

PR #125, opened 2020-11-30
Smesher api refactor

avive:


avive:
@moshababo has done most of it already. No need to merge this. Will wait for his PR.

PR #126, opened 2020-12-07
Add atx to block

avive:


avive:
The checks fail because I renamed a field which didn't use underscores between parts...

proto/spacemesh/v1/mesh_types.proto
```diff
@@ -44,7 +44,7 @@ message LayerDurationResponse {
 message MaxTransactionsPerSecondRequest {}
 
 message MaxTransactionsPerSecondResponse {
-    SimpleInt maxtxpersecond = 1;
+    SimpleInt max_txs_per_second = 1;
```
lrettig, line 47: I'd rather not make superficial breaking changes like this, since it requires updating all downstream code...

lrettig (approved):
LGTM modulo one comment

proto/spacemesh/v1/mesh_types.proto
avive: Better now than later... we just broke our own name convention here with naming this member. right?

PR #133, opened 2021-03-06
Remove activations from Layer

### LINKED ISSUE #132: Layer activations
Exactly what should `Layer.activations` contain? Should it be the activations for the current layer, or the _active set_ for the current layer (which would be activations from previous layers)?

https://github.com/spacemeshos/api/blob/105249951c66561cfc52195433ecae9cd5a121ff/proto/spacemesh/v1/types.proto#L120

lrettig:
Closes #132

Motivation: #132, https://github.com/spacemeshos/go-spacemesh/issues/2314

Activations for a given layer is not well-defined since activations are associated with blocks and epochs, not layers. Remove this to prevent confusion. They can be read from blocks for now.

### DIFF
--- a/proto/spacemesh/v1/types.proto
+++ b/proto/spacemesh/v1/types.proto
@@ -117,8 +117,7 @@ message Layer {
     LayerStatus status = 2;
     bytes hash = 3; // computer layer hash - do we need this?
     repeated Block blocks = 4; // layer's blocks
-    repeated Activation activations = 5; // list of layer's activations
-    bytes root_state_hash = 6; // when available - the root state hash of global state in this layer
+    bytes root_state_hash = 5; // when available - the root state hash of global state in this layer
 }
 
 message LayerNumber {


avive:
Why shouldn't the api support getting activations form layer which is just an aggregation of all activations in blocks of this layer? Very convenient. Without this clients have to enumerate all blocks in that layer and pull the activations from them...

avive:
As far as I remember, there was actually an ask in the api design review from research, to remove the notion of activations from a block as they are free-standing and not really belong to a block. In case they are still free-standing, the api must provide access them by layer as this is their context. @noamnelke . Activations must have context in the protocol - block, layer or epoch - so the api should support fetching them by that context....

lrettig:
> Why shouldn't the api support getting activations form layer which is just an aggregation of all activations in blocks of this layer?

Because there is no meaningful, logical association in our data model between activations and layers. This is just a convenience, or shortcut. I think the API should accurately present the data model. I don't think it's too much extra work to expect clients to download blocks and read activations from those blocks.

lrettig:
> the api must provide access them by layer as this is their context

The context of an activation is an epoch, not a layer. They are published in epoch N to guarantee eligibility in epoch N+1.

noamnelke:
>The context of an activation is an epoch, not a layer.

This was also my understanding (and maybe I convinced you?) until last Monday's research meeting, where Tal convinced me that, for security reasons, ATXs must be more specifically positioned in time (i.e. in a layer, not an epoch).

Does this need to be exposed in the API? I'm not sure, but there's also no strong reason not to expose this. If for product reasons it makes more sense to associate ATXs with epochs (I actually think it does, since it's harder for users to understand the security reasons they need to be associated with a layer) - then the presentation layer can aggregate ATXs from all layers in an epoch into one list.

avive:
From a product perspective it makes more sense to associate ATXs with layers because it gives smesher better resolution of when their ATX got on the mesh but it is not a huge deal. It is a kind of an assurance and confirmation when you see your submitted ATX in explorer and the data is less abstract because a layer time resolution is higher than an epoch.

lrettig:
But where do ATXs go in our (internal) data model? Right now, an epoch consists of a set of layers, each of which contains blocks, each of which contains transactions. ATXs are current free-standing, but point to a particular epoch. Where will the association between ATX and layer live?

noamnelke:
>but point to a particular epoch

They point to a layer, we just never implemented any validation that it was actually published in time. Each ATX has a `PubLayerID` (part of the `NIPSTChallenge` with is, in turn, part of the `ActivationTxHeader`).

lrettig:
Okay. If we're certain of this course of action, then we can close this PR. We can probably also close the linked issue in this case, since there's no change to the API design necessary (just change to the implementation).

Issue #139, opened 2021-03-17
Feature request: Node's Activation Data


avive:
# Motivation
We'd like to make the non-trivial activations flow transparent to users.
Some activation state is local to the node which submits them to PoET and to the mesh and can't be obtained from the MeshService. We should add to the NodeService capabilities to provide information about locally created and submitted activations.

## Activations Local State
1. Submitted to PoET service by node.
2. Got PoET result to node.
3. Submitted to the mesh by node.

Each state has a time-stamp.

## Proposed new functionality 
- Add method to NodeService for client to query data about all locally created and submitted activations by the node.
- Add a streaming NodeService method for clients to get a callback when a locally submitted activation state changes.

avive:
So we have basic activation data for clients via the mesh api. 

What we need to do to provide a decent UX for activations, is to add additional meta-data to activations returned by the api so clients have sufficient data to provide useful information to users and to keep them informed and on top of their activations. Here's an initial proposal for api additions:

An activation can only be in one of these states. 

```proto
enum State {
	STATE_CREATED = 0;  // from post data
	STATE_SUBMITTED = 1; // to poet
	STATE_GOT_PROOF =2;  // from poet
	STATE_BROADCASTED = 3;  // to p2p net
	STATE_MESH =4;  // published on mesh
}
```

Add `ActivationExtraData` field to the `Activation` protobufs data type or add these fields directly to `Activation` if can be done in backwards compatible manner.

```proto
// timestamps are 0 when they are not set
message ActivationExtraData {
        uint64 Submitted =1; // timestamp when submitted to poet
	uint64 Proof = 2; // proof time  
	uint64 Broadcast = 3; // broadcast time
	uint64 MeshEpoch = 4; // epoch published on mesh (eligibility epoch is this epoch + 1 )
}

```

avive:
@noamnelke @lrettig @moshababo fyi

avive:
@moshababo - what we talked about - please review.

PR #145, opened 2021-05-24
New Smesher API

moshababo:
Main remarks:

* Most methods which were directly dealing with the PoST setup were removed.
* Changed `IsSmeshing() -> bool` with `SmeshingStatus() -> [IDLE, CREATING_POST_DATA, ACTIVE]`. Since Smeshing now triggers the PoST setup, we have an intermediate stage in which we are neither idle or Smeshing (NOTE: `ACTIVE` is probably a bad name because it has nothing to do with the activation status of the miner). 
* As previously discussed, `PostStatus()` query is no longer supported (because a status can be evaluated only in relation to a`PostInitOpts` arg), besides within `PostDataCreationProgressStream`, which is supported only when a PoST data creation session is active. For this reason I changed the status message name to `SessionStatus` and removed most of its fields (`files_status`, `init_in_progress`, `error_type`, `error_message`), while keeping only `num_labels_written` (instead of `bytes_written`). We can rethink this and decide whether we want to expose errors within the stream or elsewhere, asynchronously.
* Within `PostInitOpts` the user is now to select its commitment size via `num_units` instead of `data_size`. For evaluating the size in bytes, clients are expected to query `Config()` which specifies `labels_per_unit` and `bits_per_label`.
* `PostComputeProviders()` now have `bool benchmark`, to avoid imposing the benchmark procedure delay by default (for example when a client wants to enumerate the providers for ID validation purposes).


proto/spacemesh/v1/smesher_types.proto
```diff
@@ -4,8 +4,14 @@ option go_package = "github.com/spacemeshos/api/release/go/spacemesh/v1";
 import "spacemesh/v1/types.proto";
 import "google/rpc/status.proto";
 
-message IsSmeshingResponse {
-    bool is_smeshing = 1;
+message SmeshingStatusResponse {
```
avive, line 7: I would change this method name to indicate its meaning - the post init setup status. The broader `smeshing` status is dependent on activations related flow. Please see #139 regarding providing data for the user via additional Activation data object data items. So, this method should be called something more concrete like `PosInitStatus()`.

proto/spacemesh/v1/smesher_types.proto
```diff
@@ -4,8 +4,14 @@ option go_package = "github.com/spacemeshos/api/release/go/spacemesh/v1";
...
+    enum SmeshingStatus {
```
avive, line 8: I would change this to PosInitStatus. POS = proof of space without the time part.

proto/spacemesh/v1/smesher_types.proto
```diff
@@ -4,8 +4,14 @@ option go_package = "github.com/spacemeshos/api/release/go/spacemesh/v1";
...
+        SMESHING_STATUS_UNSPECIFIED = 0; // Lane's favorite impossible value
+        SMESHING_STATUS_IDLE = 1;
```
avive, line 10: I would change the statuses to be more descriptive of the init process flow, like this:
POS_STATUS_NOT_STARTED = 1; // pos init was never executed for this node.
POS_STATUS_IN_PROGRESS = 2; // pos init (labels+pow computation) is in progress.
POS_STATUS_DONE = 3; // pos init is done and is pos data is valid
POS_STATUS_ERROR = 4; // last pos init task failed - user should call trigger init again.

avive (changes requested):

proto/spacemesh/v1/smesher_types.proto
moshababo: It’s currently covers both the post init setup status and the Smeshing boolean status.
We previously had `PostStatus()`. Are you suggesting to return it on a limited scope, in addition to having the Smeshing boolean status method?

proto/spacemesh/v1/smesher_types.proto
moshababo: I’m not sure that using PoS instead of PoST is correct. What we’re doing here is PoST init, not PoS.

moshababo: I saw that you used this abbreviation in other places as well.

proto/spacemesh/v1/smesher_types.proto
moshababo:
```
POS_STATUS_NOT_STARTED = 1; // pos init was never executed for this node.
```
We can specify that the status is idle, but not that it was never started, because we have decided to not persist the post init options, and this indication can be determined only against these options.

avive: Can we tall if last post setup ended without or with an error? Or not even that?

avive: From a user-perspective, the mode detailed the status os - the better. Users want to know what happened to a pos init they started - did it fail? did it complete without an error? Is it in progress? The goal of this api is to provide them as much info about the init process as possible. If more info on pos setup needs to be persisted then I think it will be a great feature to have.

proto/spacemesh/v1/smesher_types.proto
avive: I disagree - the time element is only added by the poet interactions after the initial space is setup. It is much simpler to call the first phase (init, space commitment) proof of space setup and everything else under the 'smeshing' umbrella term which includes the time dimension involving the poet interactions, etc... What i like to communicate to users and let them understand that what they do is first commit space from their drive and next the protocol turns it into proofs of space time which gives them eligibility in different epochs... we have lots of more work on good naming of things in this area but using 'pos' and 'smeshing' is a good start imo...

proto/spacemesh/v1/smesher_types.proto
tal-m: Please call it PoST and not PoS. It's confusing enough that there are two different definitions both called "proofs of space" and also "proofs of space and time" in addition to "proofs of spacetime" (that we're using); let's stick with the terminology as published in the literature...

avive: If we stay with GetSmeshingStatus() as the name of this method then here are the responses based on my discussion with @moshababo on them:

SMESHING_STATUS_NOT_SETUP = 1;  // node doesn't have a valid pos config to use for setup
SMESHING_STATUS_CREATING_POST_DATA = 2; // pos data init process is in progress
SMESHING_STATUS_POST_DATA_CREATED = 3; // node has a valid pos to use for activations and has finished creating data (init process done)
SMESHING_STATUS_ERROR = 4; // node has pos config info (via config or flags) but pos init has failed.

- The statuses stream should return error or progress (onOf pattern).
- There should be a new method to query the node for pos data init errors. e.g. GetLastError() - it should return status code and message using grpc built-in status in responses.
- It might be good to change GetSmeshingStatus() to GetPosStatus() as the status is only about the proof of space data init  part and not activations or poet related statuses - this data will come from activations meta-data.

moshababo: * “Proof of space” is incorrect, because this phase isn’t about producing a proof. 
* “Proof of space setup” sound like you are referring to a different cryptographic primitive.
* “Proofs of space-time setup” (abbreviated “post-setup”) is a bit more lengthy, but the most accurate and least confusing IMO. Because it’s referring to a “setup”, I don’t see a problem with having the time dimension there. Removing it doesn't make it accurate, and could create more confusion, because users are expected to be familiar with the notion of space-time and the PoST abbreviation, as the protocol’s base cryptographic primitive.

proto/spacemesh/v1/smesher_types.proto
moshababo: Continued discussion: https://github.com/spacemeshos/api/pull/145#discussion_r643439432

moshababo:
Changes:

* Returned `SmeshingStatus() -> [IDLE, CREATING_POST_DATA, ACTIVE]` to `IsSmeshing() -> bool`. 
   It’s referring solely to the phase in which post setup completed and ATX creation started.
* Returned `PostStatus()` on a limited-scope. If setup started, it’s referring to the current/latest `PostInitOpts`, otherwise to _none_. Its response is identical to `PostDataCreationProgressStream`'s.
* The error in `PostStatus()` is not mutually-exclusive with other fields, so I did not use the `Oneof` protobuf pattern. 
  It can be made mutually-exclusive in `PostDataCreationProgressStream`, but I thought it’s not necessary, and keeping it identical to `PostStatus()`'s response is preferred.


We need to conclude the [terminology dicsussion](https://github.com/spacemeshos/api/pull/145#discussion_r637745566) so I could finalize the documentation and how we name things. Currently, “pos”, “pos setup”, “post data creation”, “post init”, “post setup” are used interchangeably.  I vote for the later.

avive:
> We need to conclude the [terminology dicsussion](https://github.com/spacemeshos/api/pull/145#discussion_r637745566) so I could finalize the documentation and how we name things. Currently, “pos”, “pos setup”, “post data creation”, “post init”, “post setup” are used interchangeably. I vote for the later.

For now I see you call its Post and not Pos on the API level. We can use `post setup` in stead of `pos setup` on the ui level as well to make you and tal happy.

avive:
> Changes:

Q: are you updating cli-wallet to work with these changes as well?

avive:
Can you please remove the status field from all smesher api responses? They are not needed as a grcp api response always have a built-in a status code and field.

avive:
> Changes:
> 
> * Returned `SmeshingStatus() -> [IDLE, CREATING_POST_DATA, ACTIVE]` to `IsSmeshing() -> bool`.
>   It’s referring solely to the phase in which post setup completed and ATX creation started.
> * Returned `PostStatus()` on a limited-scope. If setup started, it’s referring to the current/latest `PostInitOpts`, otherwise to _none_. Its response is identical to `PostDataCreationProgressStream`'s.

LGTM but it will be nice ot add these notes to the doc comments of the methods in smesher.proto so the intent is clearer there for all users and this will be used as base for higher-level api docs.

moshababo:
> Q: are you updating cli-wallet to work with these changes as well?

Updated here: https://github.com/spacemeshos/cli-wallet/pull/23#issuecomment-855257003. 

> Can you please remove the status field from all smesher api responses? They are not needed as a grcp api response always has built-in status code and message fields.

This pattern is used in other services as well, so I'd rather keep it following the general API design, and discuss it with @lrettig.

> LGTM but it will be nice to add these notes to the doc comments of the methods in smesher.proto so the intent is clearer there for all users and this will be used as base for higher-level api docs.

I planned to finish the docs once the design and the implementation is accepted.

proto/spacemesh/v1/smesher_types.proto
lrettig: I like "post-setup"

proto/spacemesh/v1/smesher_types.proto
```diff
@@ -77,49 +69,42 @@ enum ComputeApiClass {
...
-    bool   throttle = 4; // Throttle down setup phase computations while user is interactive on system
-    uint32 provider_id = 5; // A PostProvider id
+// Post init options. Used to define the requested options and by
+// other messages which contain the options.
+message PostInitOpts {
+    string data_dir = 1; // User provided path to create the post data files at
+    uint32 num_units = 2; // Number of PoST data commitment units to generate
+    uint32 num_files = 3; // Number of files to equally distribute the labels among
+    uint32 compute_provider_id = 4; // A PostProvider id
+    bool throttle = 5; // Throttle down setup phase computations while user is interactive on system
```
lrettig, line 79: Couldn't this be an int, to allow more fine-tuned throttling? E.g., max CPU percentage?

proto/spacemesh/v1/smesher_types.proto
```diff
@@ -77,49 +69,42 @@ enum ComputeApiClass {
...
 
 // Param passed to methods to indicate a request to delete data files
 message StopSmeshingRequest {
     bool delete_files = 1;
 }
 
-message StopPostDataCreationSessionRequest {
-    bool delete_files = 1;
+message PostComputeProvidersRequest {
+    bool benchmark = 1;
```
lrettig, line 92: What does benchmark mean here?

lrettig:
> > Can you please remove the status field from all smesher api responses? They are not needed as a grcp api response always has built-in status code and message fields.
> 
> This pattern is used in other services as well, so I'd rather keep it following the general API design, and discuss it with @lrettig.

We use this pattern for responses that would otherwise be empty. It's a way to explicitly specify success/failure.

avive:
> > > Can you please remove the status field from all smesher api responses? They are not needed as a grcp api response always has built-in status code and message fields.
> > 
> > 
> > This pattern is used in other services as well, so I'd rather keep it following the general API design, and discuss it with @lrettig.
> 
> We use this pattern for responses that would otherwise be empty. It's a way to explicitly specify success/failure.

@lrettig  - didn't you just comment on another issue that we should get rid of these statuses in empty responses? There you agreed this is unnecessary. In the current pattern - 2 distinct statuses are returned for these methods - the built in one (which is always returned) and the explicit one - not a great way to return results to api users.

lrettig:
Yes, I agree, I think we should make this change

proto/spacemesh/v1/smesher.proto
```diff
@@ -6,60 +6,50 @@ import "spacemesh/v1/smesher_types.proto";
...
 // Readonly basic node data
 service SmesherService {
-    /////// Smeshing (mining) management
-
     // Returns true iff node is currently smeshing
-    rpc IsSmeshing(google.protobuf.Empty) returns (IsSmeshingResponse);
+    rpc IsSmeshing (google.protobuf.Empty) returns (IsSmeshingResponse);
+
+    // Start smeshing. If the post data is incomplete or missing, data creation
+    // session will be preceded. Changing of the post potions (e.g., number of labels),
```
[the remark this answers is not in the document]
moshababo: It means that if the data is incomplete or missing, data creation session will occur and must complete before smeshing can start. Feel free to fix my english, but I plan to revise all the documentation anyway.

proto/spacemesh/v1/smesher_types.proto
moshababo: It's referring to the utilization of GPU cores, not CPU time. The API complies with the internal API of `gpu-post` lib, so we should consider changing it there first.

moshababo: It's referring to the utilization of GPU cores, not CPU time. The API complies with the internal API of [`gpu-post`](https://github.com/spacemeshos/gpu-post) lib, so we should consider changing it there first.

proto/spacemesh/v1/smesher_types.proto
moshababo: Request to run a short benchmarking session for each provider, so that the `performance` field of `PoSTSetupComputeProvider` will be assigned with the num of hashes per sec.

Documentation is still missing and will be complete soon.

proto/spacemesh/v1/smesher_types.proto
avive: For the release of gpu-post lib for post it is indeed a boolean and it is not going to change to an int. Let's consider this feature for next release. We need to finalize and ship and we shouldn't change features so late in the cycle.

avive:
The smesher api generally looks fine. 
I have 2 relatively minor comments.
1. there's a bit of duplication that can go away between `PostInitOpts` and `ConfigResponse` data items. Consider having a share object for post config params that is returned in ConfigResponse and is passed by caller of `CreatePostDataRequest`.
2. I would make this commit breaking and get rid of all the responses that have only status in them, e.g. `StopSmeshingResponse` - the sooner we do this the better and it is a non-backward compatible change anyhow.

moshababo:
> 1. there's a bit of duplication that can go away between `PostInitOpts` and `ConfigResponse` data items. Consider having a shareג object for post config params that is returned in ConfigResponse and is also passed by caller of `CreatePostDataRequest`.

The naming you specified are outdated, so plz use the latest. After some revisions, we now have 2 distinct configurations:
*  `PoSTConfig` (currently defined in `PoSTConfigResponse`) — protocol params only, defined via the node configuration. That’s why it's read-only.
* `PoSTSetupOpts` — params for the local PoST setup (triggered only via `StartSmeshing`.

Considering this, you suggestion doesn’t make sense to me, so let me know if I missed something.

> 2. I would make this commit breaking and get rid of all the responses that have only status in them, e.g. `StopSmeshingResponse` - the sooner we do this the better and it is a non-backward compatible change anyhow.

I'd like to do it in a follow-up PR, because it doesn't involve solely the Smesher API.

avive:
> Considering this, you suggestion doesn’t make sense to me, so let me know if I missed something.
Ok, this makes sense for both items.

proto/spacemesh/v1/smesher_types.proto
```diff
@@ -58,86 +57,74 @@ message CoinbaseResponse {
...
-        ERROR_TYPE_READ_ERROR = 2; // Failure to read from a data file
-        ERROR_TYPE_WRITE_ERROR = 3; // Failure to write to a data file
+message PoSTSetupComputeProvider {
+    uint32 id = 1; // 0, 1, 2...
+    string model = 2; // e.g. Nvidia GTX 2700
+    enum ComputeApiClass {
+        COMPUTE_API_CLASS_UNSPECIFIED = 0;
+        COMPUTE_API_CLASS_CPU = 1; // useful for testing on systems without a cuda or vulkan GPU
+        COMPUTE_API_CLASS_CUDA = 2;
+        COMPUTE_API_CLASS_VULKAN = 3;
```
lrettig, line 103: Should we maybe add a `COMPUTE_API_CLASS_OTHER`?

proto/spacemesh/v1/smesher_types.proto
```diff
@@ -58,86 +57,74 @@ message CoinbaseResponse {
...
+message PoSTSetupStatus {
+    enum State {
+        STATE_UNSPECIFIED = 0; // Lane's favorite impossible value
+        STATE_NOT_STARTED = 1; // Setup not started
+        STATE_IN_PROGRESS = 2; // Setup in progress
+        STATE_COMPLETE = 3; // Setup is complete
+        STATE_ERROR = 4; // Setup last session ended with an error
+    }
+    State state = 1;
+    uint64 num_labels_written = 4; // Number of labels (hashes) written to the data files
```
lrettig, line 127: Why are these out of order?

proto/spacemesh/v1/smesher_types.proto
avive: For what reason? These value correspond to the type of providers supported by the gpu-post lib. If/when we'll add new supported types of providers to gpu-post then we'll add them to the enum in future api releases. Don't see how other is useful here.

lrettig (commented):
I'm definitely not a fan of the `PoST` capitalization, e.g., `PoSTConfig`. I think `PostConfig` is much easier and less error prone.

avive:
> I'm definitely not a fan of the `PoST` capitalization, e.g., `PoSTConfig`. I think `PostConfig` is much easier and less error prone.

Same with me - we don't use this naming pattern anywhere in the Api and I agree that `Post` and `PostConfig` are better.

proto/spacemesh/v1/smesher_types.proto
moshababo: fixed: https://github.com/spacemeshos/api/pull/145/commits/3679ef8d0c03690f580ff23e6db11855cc223af8

moshababo:
> > I'm definitely not a fan of the `PoST` capitalization, e.g., `PoSTConfig`. I think `PostConfig` is much easier and less error prone.

> Same with me - we don't use this naming pattern anywhere in the Api and I agree that `Post` and `PostConfig` are better.

Fixed (https://github.com/spacemeshos/api/pull/145/commits/7e631f5a76d52e8bef9c26724d5a2e7bfe94e587, and in the implementation as well).

avive (approved):
LGTM

avive (approved):
As requested, dealing with empty GRPC responses will be done in another PR.

PR #147, opened 2021-05-28
adds publicKey grpc endpoint

### LINKED ISSUE #141: API to get Public Key of Node
## Description

Currently we don't have an API to get the public key of node. We need this to construct the `bootnodes` config for the nodes in our DevOps tools. Currently our DevOps tools either get this from the logs which looks something like this:

```
{"L":"INFO","T":"2021-03-30T18:00:38.417Z","N":"ec468.p2p          ","M":"local node identity","node_id":"ec468458ada0fa621c07563dc412c5e7c31979e9d630867301b42540ea35f97c","key":"23gfU2ukQpMbC7NaxgFVaCQpLP9XSNbfunp62na9QwcJ"}
```

Or read `id.json` file in the container directory:

```
/root/spacemesh/{networkID}/p2p/nodes
```

benbot:
I was looking for this myself when I came across #141.

If this looks good to everyone I could add it to go-spacemesh as well.

I was running into a weird issue when running `make build`.

It can't find the definition for `HttpRule`

Closes #141


lrettig:
I'm okay with this. @narayanprusty @AmitShaul does it look good to you guys?

narayanprusty (approved):

avive:
q: is the public key returned here different or same to smesher id in atxs?

proto/spacemesh/v1/debug.proto
```diff
@@ -13,4 +13,7 @@ service DebugService {
     // This includes each account's address, nonce and balance but excludes projection of account state.
     rpc Accounts (google.protobuf.Empty) returns (AccountsResponse);
 
+    // PublicKey returns this node's public key
```
lrettig, line 16: Could you expand this comment to explain exactly which public key is returned? go-spacemesh actually uses several public keys. See https://github.com/spacemeshos/go-spacemesh/issues/2269 for more.

lrettig (changes requested):

PR #149, opened 2021-06-07
API 2.0

avive:
Draft PR to review and discuss GlobalState service changes to support Spacemesh Apps in release 0.3.


proto/spacemesh/v1/app_types.proto
```diff
@@ -0,0 +1,124 @@
...
+    repeated bytes data = 2; // One item - primitive var data. More than one item - an array of primitive data.
+}
+
+// The address of an app template in the templates global state store
+message AppTemplateAddress {
+    bytes address = 1;
+}
+
+// Information about an app template in the global state store
+message AppTemplate {
```
YaronWittenstein, line 42: I think we can make it `Template`

proto/spacemesh/v1/app_types.proto
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// The meta-data and value of an app's variable
+message Variable {
+    VariableMetadata meta_data = 1;
+    repeated bytes data = 2; // One item - primitive var data. More than one item - an array of primitive data.
+}
+
+// The address of an app template in the templates global state store
+message AppTemplateAddress {
```
YaronWittenstein, line 37: In the SVM code there is no `AppTemplateXXX` - so I'd rename it to `TemplateAddress`

proto/spacemesh/v1/app_types.proto
```diff
@@ -0,0 +1,124 @@
...
+    VARIABLE_TYPE_AMOUNT = 10;
+    VARIABLE_TYPE_ADDRESS = 11;
+}
+
+// Variable metadata
+message VariableMetadata {
+    uint32 id = 1; // var index in app's state
+    string name = 2; // var name
+    VariableType type = 3; // type
+    bool is_array = 4; // when true, variable is an array of 0 or more primitive value types, otherwise, a single primitive type value.
```
YaronWittenstein, line 27: I'm not well versed in Profobuf - but if it's possible to specify that a variable is either a primitive or an Array of primitives I would do that.

proto/spacemesh/v1/global_state.proto
```diff
@@ -2,14 +2,28 @@ syntax = "proto3";
...
+    rpc GetAppTemplate(GetAppTemplateRequest) returns (GetAppTemplateResponse);
+
+    // Get all app templates
+    rpc GetAllAppTemplates(GetAllAppTemplatesRequest) returns (GetAllAppTemplatesResponse);
+
+    // Get app metadata and optionally current state
+    rpc GetApp(GetAppRequest) returns (GetAppResponse);
+
+    // Get an app's variable metadata and optionally value
+    rpc GetAppVariable(GetAppVariableRequest) returns (GetAppVariableResponse);
```
YaronWittenstein, line 25: As I've said - I wouldn't add it right now.
Probably it won't be used when we have only fixed-sized variables.

proto/spacemesh/v1/global_state.proto
```diff
@@ -2,14 +2,28 @@ syntax = "proto3";
...
 // Readonly global state data - current and historical.
 // Global state data is data which is not explicitly stored in the mesh.
 // Global state is modified only by the state transition function.
 service GlobalStateService {
+
     // Latest computed global state - layer and its root hash
     rpc GlobalStateHash(GlobalStateHashRequest) returns (GlobalStateHashResponse);
 
+    // Get an app template meta-data and code
+    rpc GetAppTemplate(GetAppTemplateRequest) returns (GetAppTemplateResponse);
```
YaronWittenstein, line 16: `GetTemplate` / `GetTemplateResponse` ?

proto/spacemesh/v1/global_state.proto
```diff
@@ -2,14 +2,28 @@ syntax = "proto3";
...
+
+    // Get all app templates
+    rpc GetAllAppTemplates(GetAllAppTemplatesRequest) returns (GetAllAppTemplatesResponse);
```
YaronWittenstein, line 19: Since we want to support pagination here, I prefer to call it `GetTemplates`, or maybe `TemplateIterator`

for the response type - how about `GetTemplatesResponse` ?

proto/spacemesh/v1/app_types.proto
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// A request to return a template by address from the template global state store
+message GetAppTemplateRequest {
+    AppTemplateAddress address = 1; // template address
+    bool include_code = 2; // when true, template result will include full wasm code
+}
+
+// A response with AppTemplate data
+message GetAppTemplateResponse {
```
YaronWittenstein, line 79: `GetTemplateResponse` ?

proto/spacemesh/v1/app_types.proto
```diff
@@ -0,0 +1,124 @@
...
+    string name = 2; // short descriptive name
+    AppTemplateAddress template_address = 3; // App's template address
+    LayerNumber spawn_layer = 4; // Layer in which app was spawned
+    TransactionId spawn_tx_id = 5; // Transaction which spawned the app. empty for genesis apps.
+    AccountId spawner = 6; // App spawner account
+    repeated Variable variables = 8; // ordered array of app vars and their values
+}
+
+// A request to return a template by address from the template global state store
+message GetAppTemplateRequest {
```
YaronWittenstein, line 73: `GetTemplateResquest` ?

proto/spacemesh/v1/app_types.proto
```diff
@@ -0,0 +1,124 @@
...
+// A response with variable meta-data and value
+message GetAppVariableResponse {
+    Variable variable_value = 1;
+}
+
+// A request to get a deployed app metadata and optionally all variable values as of a specific layer or latest state.
+message GetAppRequest {
+    AccountId address = 1; // app address
+    LayerNumber layer_number = 2; // layer for state. Ignored when include_var_values is set to false. pass MAX_UI32 to get latest state.
+    bool include_var_values = 3; // when true, var values will be included in response, false otherwise.
```
YaronWittenstein, line 112: I didn't add that option under the issue - but I do think it makes sense to have it

proto/spacemesh/v1/app_types.proto
YaronWittenstein: In the future, there might be more ways to represent non-primitive variables...

YaronWittenstein (changes requested):

proto/spacemesh/v1/app_types.proto
avive: Done - and same everywhere

proto/spacemesh/v1/app_types.proto
avive: done

proto/spacemesh/v1/global_state.proto
avive: let's talk about - i think it will be very useful for api clients even with fixed-size vars

proto/spacemesh/v1/app_types.proto
avive: done

proto/spacemesh/v1/app_types.proto
```diff
@@ -0,0 +1,129 @@
...
+    bool is_array = 4; // when true, variable is an array of 0 or more primitive value types, otherwise, a single primitive type value.
+}
+
+message ValuesArray {
+    repeated bytes values = 1;
+}
+
+// The meta-data and value of an app's variable
+message Variable {
+    VariableMetadata meta_data = 1;
```
lrettig, line 37:
```suggestion
    VariableMetadata metadata = 1;
```

proto/spacemesh/v1/app_types.proto
```diff
@@ -0,0 +1,129 @@
...
+    string name = 2; // var name
+    VariableType type = 3; // type
+    bool is_array = 4; // when true, variable is an array of 0 or more primitive value types, otherwise, a single primitive type value.
+}
+
+message ValuesArray {
+    repeated bytes values = 1;
+}
+
+// The meta-data and value of an app's variable
```
lrettig, line 35:
```suggestion
// The metadata and value of a state value
```

proto/spacemesh/v1/app_types.proto
```diff
@@ -0,0 +1,129 @@
...
+message Variable {
+    VariableMetadata meta_data = 1;
+    oneof data {
+        bytes value = 2; // since bytes size is arbitrary - this will support future var types
+        ValuesArray values = 3;
+    }
+}
+
+// The address of an app template in the templates global state store
+message TemplateAddress {
```
lrettig, line 45: Why do we need a distinct type for this? Why not use the existing `Address` type?

proto/spacemesh/v1/app_types.proto
```diff
@@ -0,0 +1,129 @@
...
+    bytes address = 1;
+}
+
+// Information about an app template in the global state store
+message Template {
+    TemplateAddress address = 1; // template address
+    string name = 2; // template name
```
lrettig, line 52: See comment in issue, I don't understand why this is required

proto/spacemesh/v1/app_types.proto
```diff
@@ -0,0 +1,129 @@
...
+
+    // note - the author can be obtained from the deploy transaction but it is nice to have it here
+    AccountId author = 3; // template uploader
```
lrettig, line 55:
```suggestion
    AccountId deployer = 3; // template uploader
```

proto/spacemesh/v1/global_state_types.proto
```diff
@@ -50,22 +50,42 @@ message AccountDataQueryRequest {
     uint32 offset = 3; // return results from offset
 }
 
+message ExecutionError {
```
lrettig, line 53: We have standard error/status types already, with the same data elements, maybe we don't need to introduce another one here. See "google.golang.org/grpc/codes", "google.golang.org/grpc/status"

proto/spacemesh/v1/app_types.proto
```diff
@@ -0,0 +1,129 @@
...
+    bytes code = 6; // template wasm code
+    repeated VariableMetadata vars = 7; // template variables order and metadata
+}
+
+// A Spacemesh app
+// Note that both layer and spawner can be obtained from the transaction and transaction id for apps spawned by a transaction
+// but is convenient to return them directly as part of apps' metadata.
+message App {
+    AccountId address = 1; // App's address in global space
+    string name = 2; // short descriptive name
```
lrettig, line 69: Do we need this?

proto/spacemesh/v1/app_types.proto
```diff
@@ -0,0 +1,129 @@
...
+    TemplateAddress template_address = 3; // App's template address
+    LayerNumber spawn_layer = 4; // Layer in which app was spawned
```
lrettig, line 71: Why is this needed?

proto/spacemesh/v1/app_types.proto
```diff
@@ -0,0 +1,129 @@
...
+    bool include_code = 3; // when try templates results will include full wasm code
+}
+
+// A response includes total templates available in store and a list of zero ore more templates
+message GetTemplatesResponse {
+    uint32 total_results = 1; // total number of templates
+    repeated Template templates = 2;
+}
+
+// A request to get an app's variable by var id
```
lrettig, line 101: Instead of "variable" can we give these a more descriptive name, e.g., state entry?

proto/spacemesh/v1/app_types.proto
```diff
@@ -0,0 +1,129 @@
...
+message GetVariableRequest {
+    AccountId address = 1; // app's address
+    LayerNumber layer_number = 2; // requested value as of a layer. Pass MAX_UI32 to get latest value
```
lrettig, line 104: As noted in the issue, will only work for archival nodes! What do non-archival nodes do if this is specified? Error?

proto/spacemesh/v1/app_types.proto
```diff
@@ -0,0 +1,129 @@
...
+    uint32 id = 3; // app's var id - var's unique identifier
```
lrettig, line 105: Where does the caller look this up?

lrettig (changes requested):

proto/spacemesh/v1/app_types.proto
avive: Caller has the app's schema which includes the variable ids.

proto/spacemesh/v1/app_types.proto
avive: See my comment there - full nodes will hold state for n layers back and archive nodes since genesis but from an api level it is good to have this like this and nodes will return not_found status if they can't satisfy this specific request for a layer which is not recent.

proto/spacemesh/v1/app_types.proto
avive: Which field? These are all meta-data about an app which is useful. It is useful to know in which layer an app was deployed to the mesh although this info can be obtained from the spawn transaction - it is just convince to include some metadata that can be obtained from transactions about the app in response to this call because the intent of the caller is to get this data.

proto/spacemesh/v1/app_types.proto
avive: I think it is good to have a short app name that describes it. I think that it is something that is set by spwaner and spawn time per app.

proto/spacemesh/v1/app_types.proto
avive: I'm using the svm terminology as specified by yaron, until we discuss making any changes to this terminology with yaron.

proto/spacemesh/v1/app_types.proto
avive: I think because in current SVM design, templates are stored in their own namespace which is not just an abstracted account address. We reviewed this as far as I remember in recent svm calls. Yaron can clarify.

proto/spacemesh/v1/app_types.proto
avive: done

proto/spacemesh/v1/app_types.proto
avive: done

proto/spacemesh/v1/app_types.proto
avive: Item for design review zoom

avive:
@lrettig @YaronWittenstein - I've updated the api to reflect the design review session feedback. I still believe we should have a strongly typed GetVariable() method which is very useful for quick access to app state by dapps. Perhaps we can proceed before another review session by you guys just reviewing the most recent changes? The relent types are in app_types.proto, var_types.proto and global_state.proto

proto/spacemesh/v1/api_config.yaml
```diff
@@ -103,7 +115,7 @@ http:
       post: "/v1/smesher/createpostdata"
       body: "*"
     - selector: spacemesh.v1.SmesherService.StopPostDataCreationSession
-      post: "/v1/smesher/stoppostdatacreationsession"
+      post: "/v1/smesher/stoppost:datacreationsession"
```
lrettig, line 118: is the `:` a typo?

proto/spacemesh/v1/app_types.proto
lrettig: then, rather than introducing a new type, maybe add a `namespace` to the existing `Address` type

proto/spacemesh/v1/app_types.proto
```diff
@@ -0,0 +1,111 @@
...
+message TemplateAddress {
+    bytes address = 1;
+}
+
+// Template deploy data
+message DeployData {
+    TemplateAddress address = 1; // template address
+    // note - the author can be obtained from the deploy transaction but it is nice to have it here
+    AccountId author = 2; // template uploader
+    LayerNumber deploy_layer = 3; // layer that the template was deployed - 0 for genesis templates
```
lrettig, line 19: is this actually stored in the mesh data for a template? is it necessary for SVM? (both layer, and txid)

proto/spacemesh/v1/app_types.proto
```diff
@@ -0,0 +1,111 @@
...
+    TransactionId deploy_tx_id = 4; // the id of the deploy template transaction. Empty for genesis templates.
+}
+
+// Template code
+message Code {
+    string target_svm_version = 4; // at deploy time
```
lrettig, line 25: Can we collapse target svm version and kind down into one field? Can't it be included in the code itself as a "magic word" (first few bytes)?

proto/spacemesh/v1/app_types.proto
```diff
@@ -0,0 +1,111 @@
...
+    string kind = 5; // e.g. WASM
+    repeated bytes code = 6; // wasm code
+}
+
+// Information about a template in the global state store
+message Template {
+    DeployData deploy_data = 1; // always provided.
```
lrettig, line 32: is this included in mesh data? is it required?

proto/spacemesh/v1/app_types.proto
```diff
@@ -0,0 +1,111 @@
...
+// A response with AppTemplate data
+message GetTemplateResponse {
+    Template template = 1;
+}
+
+// A request to return all templates from the templates global state store with pagination
+message GetTemplatesRequest {
+    uint32 max_results = 1; // max number of results to return
+    uint32 offset = 2; // query offset
+    bool include_code = 3; // include code in response
```
lrettig, line 58: why not just always include both code and metadata? is it any additional work on the backend? it's not that much additional data

proto/spacemesh/v1/app_types.proto
```diff
@@ -0,0 +1,111 @@
...
+    bool include_meta_data = 4; // include metadata in response if provided by author
+}
+
+// A response includes total templates available in store and a list of zero ore more templates
```
lrettig, line 62:
```suggestion
// A response includes total templates available in store and a list of zero or more templates
```

proto/spacemesh/v1/app_types.proto
```diff
@@ -0,0 +1,111 @@
...
+///////////// App API
+
+// A Spacemesh app
+// Note that both layer and spawner can be obtained from the transaction and transaction id for apps spawned by a transaction
+// but is convenient to return them directly as part of apps' metadata.
+message App {
+    AccountId address = 1; // App's address in global space
+    string name = 2; // short descriptive name (optional)
+    TemplateAddress template_address = 3; // App's template address
+    LayerNumber spawn_layer = 4; // Layer in which app was spawned
```
lrettig, line 78: as with templates, above: are spawn layer, spawn txid, and spawner really needed for SVM/stored in mesh?

proto/spacemesh/v1/app_types.proto
lrettig: I think that works, but for the record, we cannot force full nodes to hold intermediate state for any minimum number of layers since this is extraprotocol

proto/spacemesh/v1/global_state_types.proto
```diff
@@ -50,22 +50,46 @@ message AccountDataQueryRequest {
...
+}
+
+// From dev perspective we want to enable something like the following from the higher-level lang or rust code:
+// info([enty_name], [entry_text])
+// e.g. info("sig_verification", format!("%s : %d", str1, ui64))
+message LogEntry {
+    uint32 log_level = 1; // info, debug, trace, error, etc...
+    string context = 2; // function name e.g. withdraw()
+    string name = 3; // entry name
+    string entry = 4; // entry printed data
```
lrettig, line 65: in EVM/solidity, pretty sure this is a repeated set of field names and values rather than a single blob (per context/name)

proto/spacemesh/v1/global_state_types.proto
```diff
@@ -50,22 +50,46 @@ message AccountDataQueryRequest {
...
     enum TransactionResult { // the results of STF transaction processing
         TRANSACTION_RESULT_UNSPECIFIED = 0;
         TRANSACTION_RESULT_EXECUTED = 1; // executed w/o error by the STF
         TRANSACTION_RESULT_BAD_COUNTER = 2; // unexpected transaction counter
-        TRANSACTION_RESULT_RUNTIME_EXCEPTION = 3; // app code exception
+        TRANSACTION_RESULT_RUNTIME_EXCEPTION = 3; // app code execution exception
         TRANSACTION_RESULT_INSUFFICIENT_GAS = 4; // out of gas
         TRANSACTION_RESULT_INSUFFICIENT_FUNDS = 5; // failed due to sender's insufficient funds
+        TRANSACTION_RESULT_APP_NOT_FOUND = 6; // an attempt to execute an app where there's no app at the app address
+        TRANSACTION_RESULT_FUNC_NOT_FOUND = 8; // called function not found in app
```
lrettig, line 78: is this distinguished, at the SVM level, from any other simple execution error? how/why?

proto/spacemesh/v1/global_state_types.proto
```diff
@@ -50,22 +50,46 @@ message AccountDataQueryRequest {
...
+        TRANSACTION_RESULT_INSTANTIATION_FAILED = 9; // error creating an instance from a template
```
lrettig, line 79: call this spawn rather than instantiation for consistency

proto/spacemesh/v1/global_state_types.proto
```diff
@@ -50,22 +50,46 @@ message AccountDataQueryRequest {
...
+        TRANSACTION_RESULT_DEPLOY_FAILED = 10; // error deploying a template to the mesh
```
lrettig, line 80: not sure we need to break out these two failure types for spawn/deploy since it should be clear which it is from context (tx type), no?

proto/spacemesh/v1/global_state_types.proto
```diff
@@ -50,22 +50,46 @@ message AccountDataQueryRequest {
...
+
     uint64 gas_used = 3; // gas units used by the transaction
     Amount fee = 4; // transaction fee charged for the transaction (in smidge, gas_price * gas_used)
     LayerNumber layer = 5; // the layer in which the STF processed this transaction
-    uint32 index = 6; // the index of the tx in the ordered list of txs to be executed by stf in the layer.
-    bytes svm_data = 7; // svm binary data. Decode with svm-codec
+    uint32 index = 6; // the index of the tx in the ordered list of txs to be executed by stf in the layer
+
+    bytes svm_data = 7; // SVM data returned from an SVM function. Needs to be decoded with svm-codec
+    bytes app_state = 8; // when app function executed or app deployed, the app root state hash after execution has completed
```
lrettig, line 90: please call this `state_root` or something to be clearer

proto/spacemesh/v1/types.proto
```diff
@@ -74,6 +74,7 @@ message Activation {
     AccountId coinbase = 4; // coinbase account id
     ActivationId prev_atx = 5; // previous ATX pointed to
     uint64 commitment_size = 6; // commitment size in bytes
+    EpochNumber epoch = 7; // Activation on mesh epoch number (when layer is not used)
```
lrettig, line 77: I don't understand the comment here, when is "layer not used"?

proto/spacemesh/v1/types.proto
```diff
@@ -110,9 +111,11 @@ message Block {
 message Layer {
     LayerNumber number = 1; // layer number - not hash - layer content may change
     enum LayerStatus {
-        LAYER_STATUS_UNSPECIFIED = 0; // not yet approved or confirmed
+        LAYER_STATUS_UNSPECIFIED = 0;
         LAYER_STATUS_APPROVED = 1; // approved by hare
         LAYER_STATUS_CONFIRMED = 2; // confirmed by tortoise
+        LAYER_STATUS_PENDING = 3; // not yet approved, confirmed or verified
```
lrettig, line 117: shouldn't pending come before approved and confirmed?

proto/spacemesh/v1/var_types.proto
```diff
@@ -0,0 +1,59 @@
+syntax = "proto3";
```
lrettig, line 1: per our conversation last week, I don't think any of this belongs in the L1 API. I think it belongs at the application layer, and that this data should be returned from this API as a blob of "SVM compatible data"

lrettig (changes requested):

avive:
@YaronWittenstein @lrettig - please check the code - I think I've incorporated everything we talked about and removed non-app related changes from this pr.

proto/spacemesh/v1/app_types.proto
```diff
@@ -2,47 +2,32 @@ syntax = "proto3";
...
+    repeated bytes code = 3; // wasm code
 }
 
 // Information about a template in the global state store
 message Template {
-    DeployData deploy_data = 1; // always provided.
-    Code code = 2; // template wasm code. Provided if caller specified to get code.
-    bytes metadata = 3; // storage layout, api, header. Provided if caller specified to get metadata and template include metadata.
+    Code code = 1; // template wasm code. Provided if caller specified to get code.
+    bytes metadata = 2; // storage layout, api, header. Provided if caller specified to get metadata and template include metadata.
```
lrettig, line 18:
```suggestion
    bytes metadata = 2; // storage layout, api, header. Provided if caller specified to get metadata and template including metadata. Caller is responsible for decoding.
```

proto/spacemesh/v1/app_types.proto
```diff
@@ -2,47 +2,32 @@ syntax = "proto3";
...
 }
 
 enum TemplateDataFlags {
     TEMPLATE_DATA_FLAGS_UNSPECIFIED = 0;
     TEMPLATE_DATA_FLAGS_CODE = 1;
     TEMPLATE_DATA_FLAGS_METADATA = 2;
+    // may need deploy data section - tbd - based on deploy tx format. Understand eth one.
```
lrettig, line 25:
```suggestion
    // TODO: may need deploy data section - tbd - based on deploy tx format. Understand eth one.
```

proto/spacemesh/v1/app_types.proto
lrettig: as discussed today, these also belong in metadata, as they're not required by SVM. they can optionally be automatically populated at deploy time if the deployer requests this, and pays for it.

proto/spacemesh/v1/app_types.proto
lrettig: @YaronWittenstein said these are called "storage variables" in SVM

proto/spacemesh/v1/global_state_types.proto
```diff
@@ -70,14 +70,10 @@ message TransactionReceipt {
...
-        TRANSACTION_RESULT_BAD_COUNTER = 2; // unexpected transaction counter
-        TRANSACTION_RESULT_RUNTIME_EXCEPTION = 3; // app code execution exception
-        TRANSACTION_RESULT_INSUFFICIENT_GAS = 4; // out of gas
-        TRANSACTION_RESULT_INSUFFICIENT_FUNDS = 5; // failed due to sender's insufficient funds
-        TRANSACTION_RESULT_APP_NOT_FOUND = 6; // an attempt to execute an app where there's no app at the app address
-        TRANSACTION_RESULT_FUNC_NOT_FOUND = 8; // called function not found in app
-        TRANSACTION_RESULT_INSTANTIATION_FAILED = 9; // error creating an instance from a template
-        TRANSACTION_RESULT_DEPLOY_FAILED = 10; // error deploying a template to the mesh
+        TRANSACTION_RESULT_RUNTIME_EXCEPTION = 2; // app code execution exception=
+        TRANSACTION_RESULT_VALIDATION_FAILED = 3; // TBD
```
lrettig, line 74:
```suggestion
        TRANSACTION_RESULT_VALIDATION_FAILED = 3; // TODO: do we need this? can an invalid tx ever make it into a superblock?
```

lrettig (changes requested):

avive:
Renamed pr to api 2.0 as it is not going to be backward compatible with 1.x api due to some breaking changes.

lrettig:
> Renamed pr to api 2.0 as it is not going to be backward compatible with 1.x api due to some breaking changes.

I think we need to update this so the output proto files go into `proto/spacemesh/v2` (rather than `v1`), that should cause the CI tests to pass. We also need to add a separate set of checks for `v2`, and we should also update the README accordingly.

PR #155, opened 2021-07-09
mesh: use mesh transaction in AccountMeshDataQuery rpc

### LINKED ISSUE #153: Add layer information for transactions in MeshService
# Motivation
We have a deficiency in the current api design where MeshService doesn't provide clients with the mesh layer that transactions are in. Clients need this information for: 1. order transactions in display. 2. display tx status based on tx layer and current node's notion of verified layer. e.g. `Pending` / `Confirmed`. Both are basic tx information that users of the platform will expect to be able to see. 

Note that having the tx receipt api implemented (it is currently not implemented) would not have fully resolved this issue as clients wants the above data for txs that haven't yet been executed by the STF and receipts are only generated by STF execution.

We need to solve this issue to provide a minimally usable experience for smapp 0.2 clients such as smrepl and smapp.

# Proposed solution

There are 2 main proposed solutions to address this issues for 0.2 based on the above new api type:

1. A fully backward compatible solution: Add `AccountMeshDataQueryEx` and `AccountMeshDataStreamEx` to return `AccountMeshDataEx`.

```proto
message MeshTransaction {
    Transaction transaction = 1;
    LayerNumber layernum = 2;
}

message AccountMeshDataEx {
    oneof datum {
        MeshTransaction mesh_transaction = 1;
        Activation activation = 2;
    }
}
```

2. Update AccountMeshData:

```proto
message AccountMeshData {
    oneof datum {
        MeshTransaction mesh_transaction = 1;
        Activation activation = 2;
    }
}

message MeshTransaction {
    Transaction transaction = 1;
    LayerNumber layernum = 2;
}
```

I recommend we adopt the 2nd solution as the fix for the current main client - explorer backend should be quite straightforward to apply, and long-term the api remains cleaner and simpler to use this way without ant hackish `extended` methods. Also note that the second solution makes the api more consistent because `Activation` already includes the layer number.

dshulyak:
Closes #153


avive (commented):
LGTM but maybe we should add block_id as well in MeshTransaction?

dshulyak:
Actually, I added it and then removed because it won't be so easy to add to the streaming interface. LayerID can be added to stream without significant refactoring, but BlockID is not propagated to the place where tx notification is made.

Maybe I shouldn't use MeshTransaction for streaming response.

avive:
> Maybe I shouldn't use MeshTransaction for streaming response.

I think we should because clients definitely need it in the stream as well as in the query.

lrettig:
> > Maybe I shouldn't use MeshTransaction for streaming response.
> 
> I think we should because clients definitely need it in the stream as well as in the query.

Again, we designed transaction objects in the API to not contain layer info because a transaction can exist before being added to a layer, e.g., while in the mempool. There are places in go-spacemesh where a tx event is reported with no context of layer or block.

dshulyak:
I assumed that this API is for AccountMeshData that is already included in the mesh, is this not the case?
Based on my observation, transactions are always saved only after the layer is known. And they are pushed to the notification engine  in two cases:
- if dropped from tx pool (layer is not known)
- after applying transaction to the state (layer is known)
But the streaming interface that I talked about actually ignores the first, and only publishes the second.

dshulyak:
As for receipts, are they used in the current codebase? I can't find code that pushes them to events.ReportReceipt

lrettig:
I think the correct way to solve this problem, without breaking the existing API design, is to use receipts. That's what they're for. I don't think a layer should be included in a transaction object. On the contrary, the transaction object should be included in a block, which should be included in a layer.

Receipts haven't been implemented yet. But we could probably quickly implement something very basic that just includes the required information while we finalize the design.

avive:
> I think the correct way to solve this problem, without breaking the existing API design, is to use receipts. That's what they're for. I don't think a layer should be included in a transaction object. On the contrary, the transaction object should be included in a block, which should be included in a layer.

@lrettig - receipts won't provide a solution to all cases. Clients need to know which layer a transaction they are interested at is on the mesh and such transactions can be on the mesh before a receipt for them has been created. This will be the case from the time node knows they got to a layer until the STF is executed for transactions in this layer. My working assumption for 0.2 clients has been that receipts will not be implemented for 0.2. When clients display transactions, they need to order them according to some time measure, otherwise they appear random in the transactions log. This is why we just want the layer number indicator in MeshTransaction. So, I see no reason why we should not add layer number to transactions returned by the MeshService when the transactions are in a layer in the mesh - and yes, it is a shortcut for getting the block and getting the layer for that block. User clients, in general, do not care about blocks at all - but they do care a lot of timing information for transactions they care about...

avive:
@lrettig - following our discussion about this:

TransactionState in current API release:

```proto
// TransactionState is the "journey" of a tx from mempool to block inclusion to
// mesh to STF processing. To know whether or not the tx actually succeeded,
// and its side effects, check the Receipt in the GlobalStateService.
message TransactionState {
    TransactionId id = 1;
    enum TransactionState {
        TRANSACTION_STATE_UNSPECIFIED = 0; // default state
        TRANSACTION_STATE_REJECTED = 1; // rejected from mempool due to, e.g., invalid syntax
        TRANSACTION_STATE_INSUFFICIENT_FUNDS = 2; // rejected from mempool by funds check
        TRANSACTION_STATE_CONFLICTING = 3; // rejected from mempool due to conflicting counter
        TRANSACTION_STATE_MEMPOOL = 4; // in mempool but not on the mesh yet
        TRANSACTION_STATE_MESH = 5; // submitted to the mesh
        TRANSACTION_STATE_PROCESSED = 6; // processed by STF; check Receipt for success or failure
    }
    TransactionState state = 2;
}

```

Note that it is design to return all possible state of a tx in the system, including execution results, rejection, mehs inclusion, etc...

Here's my understanding of the txs design pattern for 0.3:
1. Call TransactionsState our subscribe to TransactionsStateStream in TransactionService to get the TransactionState of a transaction by id. The TransactionState as currently implemented provide an accurate and high-level status. For example, using it client can tell if the tx is in the mem-pool, on the mesh or was executed.
2. If tx status shows it was executed then call GlobalStateServicee to get a tx receipt with detailed info.
3. If tx is on mesh but not executed - call the MeshService to get the tx mesh status - this is where we have a gap in our understanding. It is my understanding that the response should include all the mesh-related info about the tx, including layerId and blockId.

Tx receipts are not planned to be implemented for 0.2. So for 0.2 the pattern is: 
1. Call TransactionsState our subscribe to TransactionsStateStream in TransactionService to get the TransactionState of a transaction by id. The TransactionState as currently implemented provide an accurate and high-level status. For example, using it client can tell if the tx is in the mem-pool, on the mesh or was executed.
2. If tx is on mesh but not executed - call the MeshService to get the tx mesh status - this is where we have a gap in our understanding. It is my understanding that the response should include all the mesh-related info about the tx, including layerId and blockId.
3. Use the TranactionState to display information for the client, for example TRANSACTION_STATE_PROCESSED, TRANSACTION_STATE_MESH, TRANSACTION_STATE_CONFLICTING.

Let's first agree that these patterns are accurate. If they are then it makes sense to me to add mesh-info to the MeshService tx related methods.

lrettig:
> Clients need to know which layer a tx they are interested at is on the mesh, and such txs can be on the mesh before a receipt for them has been generated. This will be the case from the time node knows a tx got to a mesh layer and until the time that STF is executed for txs in this layer.

> If tx is on mesh but not executed - call the MeshService to get the tx mesh status

This doesn't make sense to me. There is no such thing as "on the mesh but not executed." When a tx appears "on the mesh," that means it's been mined into a block and put in a layer, the layer has been confirmed, and all nodes should have already run the transaction. It's atomic. Are you suggesting something like "a miner has included the transaction in a proposed block (that hasn't yet been confirmed as part of any layer"? That's still just pending, in my mind.

There is no "from the time node knows a tx got to a mesh layer and until the time that STF is executed for txs in this layer" - again, this is atomic.

The right thing to do is to 1. use the tx service to submit the tx and get the txid, then 2. follow the `LayerStream` in the mesh service until you see the tx appear in a block in a confirmed layer. Alternatively you can 3. follow the `GlobalStateStream` in the global state service and look for a corresponding tx receipt.

We may want to add filters to one or both of these streams to make it easier to filter for a specific txid.

I'm still a little bit uncomfortable with adding layer number to the transaction data object (directly, or indirectly as a wrapper). We have a very neat directed abstraction: epoch -> layer -> block -> tx. By adding a link from tx -> layer, you add a cycle in this otherwise acyclic graph.

In design terms, I think a good API is opinionated and presents exactly one way to retrieve each unit of data (rather than lots of hacks or shortcuts). I think the right approach here is to:

1. implement tx receipts (at least very basic functionality)
2. add filters to the existing query and stream endpoints, as necessary, if smapp can't handle the volume of data (this is the existing design pattern that we chose)

avive:
Thanks for your detailed explanation. So this means that:
1. We *must* implement the transaction receipts methods in the GlobalStateService in order ship 0.2 smapp and smrepl and we can't hack around it and this becomes an 0.2 blocking issue.
1.  don't think we want or need implement filtering by txs ids in MeshService if indeed we modify the definition of a verified layer to mean that the txs on this layer have been executed because it is unnecessary and requires constant extra computation in the node when clients are registered on that stream - it will need to compare each tx in each block with a list of one or more txs registered by clients in filters.

lrettig:
My proposal for how to deal with this is as follows:

- client generates and submits a new tx using `TransactionService.SubmitTransaction` (and gets a new txid in response)
- client shows a "pending" state for this tx
- client calls `MeshService.LayersQuery` once per layer and iterates through the response data to check if the transaction was successfully added to the mesh. if so, client updates the UI with layer timestamp and marks the tx as successful/final.
- if the tx does _not_ appear in the mesh after one or two layers, mark the tx as failed and suggest that the user submit it again

Without having implemented tx receipts, we have no way to _prove_ that a tx failed, but if it doesn't appear in the mesh within one or two layers max, we know implicitly that this is what happened. This is sufficient for now. We can do this more explicitly (and provide more info on what happened, e.g., out of gas, bad nonce, VM error, whatever) later when we properly implement transactions.

I'm going to close this since this use case does not require API changes at present.

avive:
This proposal needs to be clarified a bit to consider that clients will often go offline and online again.
- After client queried a layer for txs, it needs to persist this layer number.
- When new app session begin, client needs to load the stored layer number and query the MeshService for all layers from that layer until the tip (current layer) to transactions and look for transactions of interest in these layers.
- Once a transaction is in a layer, associate layer number with a tx.

lrettig:
Discussed this with @avive further. We agreed that it's okay to add this to API v1 for convenience (until tx receipts have been implemented), but it will not be included in API v2.

lrettig (approved):

lrettig:
you can go ahead and merge - merging in this repo does not require that CI be passing, that's just informational (this is a breaking change)

Issue #144, opened 2021-05-07
Redesign layer status to match new consensus mechanisms


lrettig:
Currently a layer can have one of three statuses:

https://github.com/spacemeshos/api/blob/105249951c66561cfc52195433ecae9cd5a121ff/proto/spacemesh/v1/types.proto#L112-L116

These statuses no longer map to how our consensus mechanisms actually work. Here's a better, more accurate design:

- unspecified
- pending: layer is still syncing, or new, or otherwise waiting to be processed/validated
- analyzing: hare is currently running for this layer
- invalidated: hare failed to run for this layer, and the node has decided it's empty (all blocks are invalid); or, hare succeeded and confirmed an empty layer
- tentative: hare succeeded running for the layer, and a node thus has weak confidence that all other nodes agree on its contents
- stuck: tortoise tried to verify this layer, but couldn't (because hare hasn't finished yet, or because the global opinion on the layer is abstain and the layer isn't old enough yet to try healing)
- confirmed: tortoise succeeded in verifying the layer
- applied: layer state transitions have been applied, receipts generated, etc.
- healing: even after a layer is confirmed and its state has been applied, in rare cases, a node may need to re-apply the layer as part of a self-healing process
- final: at some point we may want to go one step further and say that the layer is totally final and its state can no longer be updated, even by self-healing (I'm not sure when or if we can say this)

We may not want to surface all of these possible statuses via the API, and this list is not precisely [MECE](https://en.wikipedia.org/wiki/MECE_principle) as there is some overlap, but it's reasonably comprehensive.

Related: https://github.com/spacemeshos/go-spacemesh/issues/2403

lrettig:
To be clear, transactions obviously do not have an independent status - they derive their status from the status of their block and layer.

> while in self-healing, no other data is canonical until the self healing is complete

What makes self-healing complex, in this context, is that it can _invalidate_ a previously valid block (or vice-versa). So we could have blocks (and transactions) that are "approved" and applied to state, then reverted later. That's why I suggested introducing a "final" status, but we'll have to discuss with @tal-m the threshold beyond which we could apply this.

avive:
We need to refine this and find a minimal MECE set. For example, why do we need `unspecified` if we have `pending`? Obviously we need to find a balance between being descriptive and informative and not confusing users with too many states. I think 7 is the magic number here that above it most people will the states just overwhelming and overly complex. For example, if stuck is a temporary possible state then it can also be pending.
[...]

PR #156, opened 2021-07-12
Update LayerStatus for 0.2 Consensus Changes

avive:
Add new layer statuses to conform with consensus protocol changes. Refactor this change from api-related changes in Api 2.0 PR.


lrettig:
I'm a bit confused about this PR. I thought we already agreed (more or less) upon an updated spec here, for API 2.0? https://github.com/spacemeshos/api/issues/144#issuecomment-841787555

Why not just go with this?

avive:
this is for spacemesh 0.2 and not for 0.3 hence a separate pr and should go into 1.x and not into 2.0. It has been a while since we discussed this but as far as i remember we reviewed this together and agrees that this reflects 0.2 consensus.

lrettig:
> this reflects 0.2 consensus

Could you refresh my memory? Is there a thread I can review?

> confirmed by tortoise
> verified by tortoise

what's the difference between these two?

avive:
The answer depends on possible layer states in the node because the API just need to reflect them. We need to get someone who's familiar with latest Tortoise implementation to chime in on this. It is my understanding that the node has a separate notions of `verified` and `confirmed` and also about layers it knows about which are not verifier or confirmed - `pending`. @noamnelke - what's the deal here?

lrettig:
I'm quite familiar with the latest Tortoise implementation :) From the perspective of Tortoise, a layer can only have two states, unverified or verified. (You could maybe posit a third, but I don't see what we gain from it: tried to verify, and failed, and will try again later.)

You may be thinking of the case where _Hare_ finishes for a layer which hasn't been verified by Tortoise yet, but this is already accounted for in the API.

avive:
> I'm quite familiar with the latest Tortoise implementation :) From the perspective of Tortoise, a layer can only have two states, unverified or verified. (You could maybe posit a third, but I don't see what we gain from it: tried to verify, and failed, and will try again later.)

Cool, so, are you 100% sure that there is no separate `verified` and `confirmed` tortoise layer states in the protocol 0.2? If not then we don't need this change, if yes, then we do.

> You may be thinking of the case where _Hare_ finishes for a layer which hasn't been verified by Tortoise yet, but this is already accounted for in the API.

If a layer is not yet approved by hare but known to node that it won't be reported by the 0.1 api. There can be a layer in this state known to node which will not be reported via the 0.1 api as it doesn't have a state for it.

lrettig:
This is not necessary for SM 0.2. Going to go ahead and close this. Let me know if I've missed anything!

PR #159, opened 2021-11-22
Do not omit some zero values in JSON

nkryuchkov:
Many JSON fields use `omitempty` tag now which makes JSON encoder omit them if they have a default value for their type. It's not an often case, but anyway. For example, if someone wants to send 0 coins, an API reply would look weird:
```
"amount": {
         
},
```

Also, JS would interpret this as `undefined` instead of `0`/`0.0`/`""`.

The PR removes `omitempty` in some places where it isn't needed


nkryuchkov:
I didn't notice that I changed the auto-generated code. The protobuf compiler adds the `omitempty` tag everywhere by default, so that's the reason. Therefore, I'm closing the PR

PR #161, opened 2022-01-14
debug: add network info response endpoint with network identity

dshulyak:
Existing tooling relies on logs or files to recover this information, i want to simplify the integration for new tooling.


nkryuchkov (approved):

lrettig:
What's meant by `NetworkInfo`? In particular shouldn't we consider the versioning information we defined in https://github.com/spacemeshos/SMIPS/issues/75? I feel like this should maybe be a struct including all of these version IDs.

dshulyak:
> What's meant by NetworkInfo? In particular shouldn't we consider the versioning information we defined in spacemeshos/SMIPS#75? I feel like this should maybe be a struct including all of these version IDs.

id is a p2p identity. network info means that this struct provides network info, it may include currently connected peers for example. i don't see how that smip is related to my use case, as it is a debug service we should have no problems update it in future

PR #167, opened 2022-06-24
3222 removed unnesessary size in response

abergasov:
for endpoint `EstimatedRewards` was added ability to set layer, for which reward should be calculated

in response was removed `num_units` field as not need

related to pr https://github.com/spacemeshos/go-spacemesh/pull/3296


proto/spacemesh/v1/smesher_types.proto
```diff
@@ -57,16 +57,16 @@ message CoinbaseResponse {
...
+}
 
 // Estimated rewards for the next epoch. Note that this is a global value that will be the
 // same for all smeshers, and is based on the default minimum commitment size and fixed
 // epoch length, both of which are network params.
 message EstimatedRewardsResponse {
     // The amount of the total estimated reward in the next upcoming epoch
     Amount amount = 1;
-    // The number of Post data commitment units that this estimated reward corresponds to (part of global config)
-    uint32 num_units = 2;
```
dshulyak, line 69: are you sure that it needs to be removed?

abergasov: `NumUnits` looks like is how much space was allocated, not looks like it really need for this endpoint.
We can add it later, if it will be need for some reasons

lrettig: Is this comment still accurate?

> Note that this is a global value that will be the
> // same for all smeshers, and is based on the default minimum commitment size and fixed
> // epoch length, both of which are network params.

PR #170, opened 2022-06-30
new endpoint for `SmesherService.SmeshingStatus`

abergasov:
adding new endpoint for getting info about requested epochs

original issue https://github.com/spacemeshos/go-spacemesh/issues/3281

example response
```
type SmeshingStatusResponse struct {
  status: 'not_eligible', 
  epoch: uint64;
  estimatedTime: number;
  reason: string;
  processedBytes: string // like "Processed 176 GB of 256 GB"
  processedPercentage: string // like "86% completed"
}
```


dshulyak:
is there an open pr? i don't quite understand some fields.

what is `status: not eligible`?  you are always eligible to smesh

proto/spacemesh/v1/smesher_types.proto
```diff
@@ -128,3 +128,17 @@ message PostSetupStatus {
     PostSetupOpts opts = 3; // setup options previously set by the user
     string error_message = 4; // The error message, if the state is STATE_ERROR
 }
+
+message SmeshingStatusRequest {
+    string smesher_id = 1;
```
dshulyak, line 133: smesher id should be bytes

i think we are never building post or atxs for more than one smesher. why do you need smesher id in the request?

proto/spacemesh/v1/smesher_types.proto
```diff
@@ -128,3 +128,17 @@ message PostSetupStatus {
...
+    repeated uint32 epochs = 2; // Epochs to query status for. If empty, returns status for current, next and subsequent.
```
dshulyak, line 134: atx is always built for next epoch

lrettig: agree, I don't think this request needs to include the epoch, it can only apply to the current/next epoch.

lrettig: I'm not even sure it needs to include the smesher_id (in other words, we may not need this message type at all) - the node it's talking to already knows its smesher ID, and probably wouldn't have data on another smesher's status, would it?

proto/spacemesh/v1/smesher_types.proto
```diff
@@ -128,3 +128,17 @@ message PostSetupStatus {
...
+}
+
+message SmeshingStatusResponse {
+    string status = 1;
+    uint32 epoch = 2;
+    uint64 estimated = 3;
+    string reason = 4;
+    string processed_bytes = 5;
```
lrettig, line 142: How is this different than `PostSetupStatus`? Which bytes are being counted here?

abergasov: this is for ui inform view. like show 
```
 processedBytes:  "Processed 176 GB of 256 GB"
 processedPercentage: "86% completed"
```

proto/spacemesh/v1/smesher_types.proto
abergasov: yes, can use nodeID instead smesherID

abergasov:
> is there an open pr? i don't quite understand some fields.
> 
> what is `status: not eligible`? you are always eligible to smesh

related pr
https://github.com/spacemeshos/go-spacemesh/pull/3314

`not eligible` is related to specific epoch X. if you not build file and not submit proff before epoch start - you `not eligible` build block and get reward for started epoch X

proto/spacemesh/v1/smesher_types.proto
abergasov: yes, mistake, can use nodeID instead, if will need

proto/spacemesh/v1/smesher_types.proto
lrettig: What do these bytes represent? I don't understand

Also the data passed from the backend to the frontend should be numeric (i.e., not strings). The frontend can interpret and display that data in different ways (e.g., using different languages)

proto/spacemesh/v1/smesher_types.proto
lrettig: I don't think we need `epochs` here either. This endpoint doesn't need to take any args, and it can just return status for the current and next epochs.

proto/spacemesh/v1/smesher_types.proto
```diff
@@ -128,3 +128,16 @@ message PostSetupStatus {
     PostSetupOpts opts = 3; // setup options previously set by the user
     string error_message = 4; // The error message, if the state is STATE_ERROR
 }
+
+message SmeshingStatusRequest {
+    repeated uint32 epochs = 1; // Epochs to query status for. If empty, returns status for current, next and subsequent.
+}
+
+message SmeshingStatusResponse {
+    string status = 1;
```
lrettig, line 137: status should also be a numeric code, not a string. `reason` can be a string (or a code).

proto/spacemesh/v1/smesher_types.proto
```diff
@@ -127,4 +127,21 @@ message PostSetupStatus {
     uint64 num_labels_written = 2; // Number of labels (hashes) written to the data files
     PostSetupOpts opts = 3; // setup options previously set by the user
     string error_message = 4; // The error message, if the state is STATE_ERROR
+    string processed = 5; // The processed size, like `176 GB`
+    string post_total_size = 6; // Total size, that should be processed, like `256 GB`
+    uint32 processed_percentage = 7; // The processed percentage, like `50`
+}
+
+message SmeshingStatusResponse {
+    enum EpochStatus {
```
lrettig, line 136: enum element names should start with the enum name, i.e., "EPOCH_STATUS_*", and first one (value 0) should be `EPOCH_STATUS_UNSPECIFIED`

proto/spacemesh/v1/smesher_types.proto
```diff
@@ -127,4 +127,21 @@ message PostSetupStatus {
...
+        POST_NOT_FOUND = 0; // No ATX and creation of post not started yet.
+        INITIALIZING_POST = 1; // When node start create post file.
+        GENERATING_POS_PROOF = 2;
```
lrettig, line 139: `POS_PROOF` or `POST_PROOF`?

proto/spacemesh/v1/smesher_types.proto
```diff
@@ -127,4 +127,21 @@ message PostSetupStatus {
...
+        WITING_FOR_POET = 3;
```
lrettig, line 140:
```suggestion
        EPOCH_STATUS_WAITING = 3;
```

proto/spacemesh/v1/smesher_types.proto
lrettig: maybe just rename this one `EPOCH_STATUS_GENERATING`

proto/spacemesh/v1/smesher_types.proto
```diff
@@ -127,4 +127,21 @@ message PostSetupStatus {
...
     string error_message = 4; // The error message, if the state is STATE_ERROR
+    string processed = 5; // The processed size, like `176 GB`
+    string post_total_size = 6; // Total size, that should be processed, like `256 GB`
+    uint32 processed_percentage = 7; // The processed percentage, like `50`
+}
+
+message SmeshingStatusResponse {
+    enum EpochStatus {
+        POST_NOT_FOUND = 0; // No ATX and creation of post not started yet.
+        INITIALIZING_POST = 1; // When node start create post file.
```
lrettig, line 138:
```suggestion
        EPOCH_STATUS_INITIALIZING = 1; // When node start create post file.
```

proto/spacemesh/v1/smesher_types.proto
lrettig:
```suggestion
        EPOCH_STATUS_GENERATING = 2;
```

proto/spacemesh/v1/smesher_types.proto
```diff
@@ -127,4 +127,21 @@ message PostSetupStatus {
     uint64 num_labels_written = 2; // Number of labels (hashes) written to the data files
     PostSetupOpts opts = 3; // setup options previously set by the user
     string error_message = 4; // The error message, if the state is STATE_ERROR
+    string processed = 5; // The processed size, like `176 GB`
+    string post_total_size = 6; // Total size, that should be processed, like `256 GB`
+    uint32 processed_percentage = 7; // The processed percentage, like `50`
+}
+
+message SmeshingStatusResponse {
```
lrettig, line 135: Please break out the rest of this message into its own message type called `SmeshingStatus`. See how `PostSetupStatus` is implemented here. Please follow the existing design.

proto/spacemesh/v1/smesher_types.proto
```diff
@@ -127,4 +127,21 @@ message PostSetupStatus {
     uint64 num_labels_written = 2; // Number of labels (hashes) written to the data files
     PostSetupOpts opts = 3; // setup options previously set by the user
     string error_message = 4; // The error message, if the state is STATE_ERROR
+    string processed = 5; // The processed size, like `176 GB`
```
lrettig, line 130: I don't think this should be string, this should be a number, probably number of bytes

proto/spacemesh/v1/smesher_types.proto
```diff
@@ -127,4 +127,21 @@ message PostSetupStatus {
...
+    string post_total_size = 6; // Total size, that should be processed, like `256 GB`
```
lrettig, line 131: This should also be a number

proto/spacemesh/v1/smesher_types.proto
```diff
@@ -127,4 +127,21 @@ message PostSetupStatus {
...
+    uint32 processed_percentage = 7; // The processed percentage, like `50`
```
lrettig, line 132: This should also be a number

proto/spacemesh/v1/smesher_types.proto
lrettig: Isn't this already included in `PostSetupOpts` as `num_units` and `num_files`?

proto/spacemesh/v1/smesher_types.proto
lrettig: And maybe instead of percentage, should be number of processed bytes

proto/spacemesh/v1/smesher_types.proto
```diff
@@ -127,4 +127,21 @@ message PostSetupStatus {
...
+    enum EpochStatus {
+        POST_NOT_FOUND = 0; // No ATX and creation of post not started yet.
+        INITIALIZING_POST = 1; // When node start create post file.
+        GENERATING_POS_PROOF = 2;
+        WITING_FOR_POET = 3;
+        SMESHING = 4;
+    }
+    EpochStatus status = 1;
+    uint32 epoch = 2;
+    uint64 estimated = 3;
```
lrettig, line 145: Estimated what?

lrettig (changes requested):
We originally talked about this endpoint referring to [both current and next epochs](https://github.com/spacemeshos/api/pull/170#discussion_r912116417). Is the intention for it to refer only to the current epoch? Are you sure it doesn't need to return info for subsequent epoch?

proto/spacemesh/v1/smesher_types.proto
```diff
@@ -127,4 +127,25 @@ message PostSetupStatus {
...
+        EPOCH_STATUS_UNSPECIFIED = 0;
+        EPOCH_STATUS_POST_NOT_FOUND = 1; // No ATX and creation of post not started yet.
+        EPOCH_STATUS_INITIALIZING = 2; // When node start create post file.
+        EPOCH_STATUS_GENERATING = 3; // Generating Proof-of-Space
+        EPOCH_STATUS_WAITING = 4; // Waiting for PoET challenge
+        EPOCH_STATUS_SMESHING = 5;
+    }
+    EpochStatus status = 1;
+    SimpleInt epoch = 2;
+    uint64 estimated_time = 3; // Estimated time to complete, in seconds.
```
lrettig, line 149: How does the node calculate this?

proto/spacemesh/v1/smesher_types.proto
```diff
@@ -127,4 +127,25 @@ message PostSetupStatus {
     uint64 num_labels_written = 2; // Number of labels (hashes) written to the data files
     PostSetupOpts opts = 3; // setup options previously set by the user
     string error_message = 4; // The error message, if the state is STATE_ERROR
+    uint64 processed_bytes = 5; // The processed size, in bytes
+    uint64 post_total_size_bytes = 6; // Total size, that should be processed
+}
+
+message SmeshingStatusResponse {
+    repeated SmeshingStatus epoch = 1;
```
lrettig, line 135:
```suggestion
    repeated SmeshingStatus status_of_epoch = 1;
```

proto/spacemesh/v1/smesher_types.proto
```diff
@@ -127,4 +127,25 @@ message PostSetupStatus {
     uint64 num_labels_written = 2; // Number of labels (hashes) written to the data files
     PostSetupOpts opts = 3; // setup options previously set by the user
     string error_message = 4; // The error message, if the state is STATE_ERROR
+    uint64 processed_bytes = 5; // The processed size, in bytes
+    uint64 post_total_size_bytes = 6; // Total size, that should be processed
```
lrettig, line 131:
```suggestion
    uint64 bytes_total = 6; // Total size, that should be processed
```

proto/spacemesh/v1/smesher_types.proto
```diff
@@ -127,4 +127,25 @@ message PostSetupStatus {
     uint64 num_labels_written = 2; // Number of labels (hashes) written to the data files
     PostSetupOpts opts = 3; // setup options previously set by the user
     string error_message = 4; // The error message, if the state is STATE_ERROR
+    uint64 processed_bytes = 5; // The processed size, in bytes
```
lrettig, line 130:
```suggestion
    uint64 bytes_processed = 5; // The processed size, in bytes
```

proto/spacemesh/v1/smesher_types.proto
lrettig: > consumer can calculate progress on his side, just get data from several endpoints

How? Which endpoints? How does client know the progress?

lrettig (changes requested):

PR #173, opened 2022-07-12
change address type from byte slice to string

abergasov:
change all address fields from []byte to string `0x123...`


lrettig:
Please check whether this also needs to change:

https://github.com/spacemeshos/api/blob/63f46156c23f26969593151aaa766bfb46504dfa/proto/spacemesh/v1/tx_types.proto#L61-L62

lrettig:
These are also addresses:

https://github.com/spacemeshos/api/blob/f896c521d3155ffa8dcc51c3ec10f3840ab4be6b/proto/spacemesh/v1/types.proto#L48-L49

proto/spacemesh/v1/tx_types.proto
```diff
@@ -59,7 +59,7 @@ message TransactionResultsRequest {
     // id is filter by transaction id.
     bytes id = 1;
     // address is a filter by account address, it could be principal or any affected address.
-    bytes address = 2;
```
dshulyak, line 62: don't change it please. the rest of the protocol should use flat type as well

proto/spacemesh/v1/tx_types.proto
```diff
@@ -82,5 +82,5 @@ message TransactionResult {
     uint64 fee = 5;
     bytes block = 6;
     uint32 layer = 7;
-    repeated bytes touched_addresses = 8;
```
dshulyak, line 85: revert

proto/spacemesh/v1/types.proto
```diff
@@ -45,8 +45,8 @@ message Activation {
 // do not include mutable data such as tx state or result.
 message Transaction {
     bytes id = 1;
-    bytes principal = 2;
-    bytes template = 3;
+    AccountId principal = 2;
+    AccountId template = 3;
```
dshulyak, line 49: i don't know if template should have an hrp. this seems a bit weird. @lrettig what do you think?

proto/spacemesh/v1/tx_types.proto
```diff
@@ -82,5 +82,5 @@ message TransactionResult {
...
+    repeated AccountId touched_addresses = 8;
```
dshulyak, line 85: this should be string

proto/spacemesh/v1/tx_types.proto
```diff
@@ -59,7 +59,7 @@ message TransactionResultsRequest {
     // id is filter by transaction id.
     bytes id = 1;
     // address is a filter by account address, it could be principal or any affected address.
-    bytes address = 2;
+    AccountId address = 2;
```
dshulyak, line 62: string

dshulyak: i deliberately avoided using AccountId since it creates multiple unnecessary levels, but you changed it back

proto/spacemesh/v1/types.proto
lrettig: It should. It's a domain-specific thing -- the same address (i.e., bytestr) can exist on many networks, but a template may only be deployed on one, and the HRP specifies the domain.

lrettig: I think we can make it optional for both `principal` and `template` since HRP is already implied by `GENESISID`, which the tx is bound to (https://github.com/spacemeshos/pm/issues/135)

dshulyak (approved):

PR #180, opened 2022-09-04
add LAYER_STATUS_APPLIED 

countvonzero:
for when block is applied for a layer

this is needed for systest where we checks for a layer is confirmed and grab the hash then.
but unfortunately, when a layer is confirmed by tortoise, the layer hash may not have been updated yet.
therefore, we need to add event when a block is applied for an layer.


spacemesh/v1/types.proto
```diff
@@ -95,6 +95,7 @@ message Layer {
         LAYER_STATUS_UNSPECIFIED = 0; // not yet approved or confirmed
         LAYER_STATUS_APPROVED = 1; // approved by hare
         LAYER_STATUS_CONFIRMED = 2; // confirmed by tortoise
```
dshulyak, line 97: you can probably remove CONFIRMED, it is useless now
i think it was working before the change where hare applied blocks immediately

dshulyak (approved):

spacemesh/v1/types.proto
countvonzero: but i think it still has value in the future. for us, this is more like the layer is finalized.

PR #191, opened 2022-11-01
Add ActivationService allowing to get ATX by ID

poszu:
Required for: https://github.com/spacemeshos/go-spacemesh/issues/3458

A node is supposed to implement and run `ActivationService` on its GRPC endpoint. Poet will use it for fetching ATXes.


spacemesh/v1/activation_types.proto
```diff
@@ -0,0 +1,13 @@
...
+package spacemesh.v1;
+option go_package = "github.com/spacemeshos/api/release/go/spacemesh/v1";
+
+import "spacemesh/v1/types.proto";
+
+message GetRequest {
+    bytes id = 1;
+}
+
+message GetResponse {
```
dshulyak, line 11: is there a benefit of wrapping Activation in GetResponse? i understand that it allows to add more fields (and maybe deprecate field 1) to GetResponse in the future, but what for

dshulyak (approved):

spacemesh/v1/activation_types.proto
poszu: The only motivation is to be able to change the contents of the response in a backwards-compatible fashion in the future.

PR #194, opened 2022-11-04
Fix package name in release/go directory

fasmat:
The package name of `tools.go` doesn't match the directory name causing `go get` to fail on any version after the file was added. This should fix the issue.


release/go/spacemesh/v1/activation.pb.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.28.1
```
poszu, line 3: Now the version is older and doesn't match the version of `google.golang.org/protobuf v1.28.1`

poszu (changes requested):

release/go/spacemesh/v1/activation.pb.go
fasmat: Unfortunately no, this file is generated by the tool `github.com/golang/protobuf/protoc-gen-go@v1.5.2` which seems to be the newest version. Before we were using the same tool in the same version but compiled it with a newer dependency than the one it was released with (v1.28.1 instead of 1.26.0).

Is there any particular reason as to why we should use a newer version of the dependency? The generated code is the same (except for this comment).

poszu: 🤯  If you install `protoc-gen-go@v1.5.2` then why does it show `protoc-gen-go v1.28.1`? 

Anyway, you are - right code is the same. I still wonder how it will break in the future because of this version divergence 😅

poszu (approved):

PR #220, opened 2023-03-22
add api support for epoch atx stream

countvonzero:
part of https://github.com/spacemeshos/go-spacemesh/issues/4089

for stand-alone binary to create bootstrap/fallback beacon and activeset

### DIFF
--- a/spacemesh/v1/mesh_types.proto
+++ b/spacemesh/v1/mesh_types.proto
@@ -99,3 +99,11 @@ message LayerStreamRequest {}
 message LayerStreamResponse {
     Layer layer = 1;
 }
+
+message EpochStreamRequest {
+    uint32 epoch = 1;
+}
+
+message EpochStreamResponse {
+    repeated ActivationId ids = 1;
+}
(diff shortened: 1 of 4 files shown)


spacemesh/v1/mesh_types.proto
```diff
@@ -99,3 +99,11 @@ message LayerStreamRequest {}
 message LayerStreamResponse {
     Layer layer = 1;
 }
+
+message EpochStreamRequest {
```
dshulyak, line 103: i thought that we already have some api to query atxs, we could use that until there is time to design new.

i wanted to have something similar to TransactionResult for every object.

```go
type AtxService interface {
     Stream(Request) 
}

type Request struct {
    EpochStart, EpochEnd uint32 // includes all in range[EpochStart, EpochEnd)
    Smesher []byte // if not empty filters by smesher
}
```

and basically similar api for every other object that can be consumed for various purposes

dshulyak (approved):
i think i would prefer to design some more generic, but if it works then we can use this

spacemesh/v1/mesh_types.proto
countvonzero: so if you look at api/node.go in go-spacemesh, there is no interface that allows the API to query all atx ids in an epoch.
what's currently available is for the caller to query full atx data by atx ids.

my assumption is also that we will rework something new for the api. current state is difficult for me to use. so this would be short-lived and replaced by v2.

countvonzero: i'd really like the primitive types to be consumed directly in v2. (u32, u64, []byte)

dshulyak: sounds fine to me

PR #230, opened 2023-04-15
Use uint32 for epoch number

fasmat:
We are keeping track of epoch and layer IDs as uint32s, this aligns our API to be in sync.


.github/workflows/ci.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       - name: lint and check build
         run: |
           # Runs the buf linter and checks for breaking changes
-          make https
+          # make https # TODO: temporarily disabled
```
fasmat, line 27: TODO: re-enable after merging

dshulyak (approved):
my preference is to use u32 without wrapper, as such things are very unlikely to change in future. but we can leave that for api v2 discussions

PR #247, opened 2023-07-04
admin: stream events relevant for user

dshulyak:


spacemesh/v1/admin_types.proto
```diff
@@ -14,3 +16,91 @@ message RecoverRequest {
...
+
+message Event {
+  // time when event occured.
+  google.protobuf.Timestamp timestamp = 1;
+  // the reason of the failure may be complex to pinpoint.
+  // for the first version we want to highlight that failure occured and defer user to logs. 
+  bool failure = 2;
+  // what role this event plays in the protocol. 
+  // might be long, should not be displayed by default, or displayed in small text.
+  string help = 3;
```
dshulyak, line 28: this should go as docstring for events below

spacemesh/v1/admin_types.proto
```diff
@@ -14,3 +16,91 @@ message RecoverRequest {
   string uri = 1;
   uint32 restore_layer = 2;
 }
+
+message Event {
+  // time when event occured.
+  google.protobuf.Timestamp timestamp = 1;
+  // the reason of the failure may be complex to pinpoint.
+  // for the first version we want to highlight that failure occured and defer user to logs. 
+  bool failure = 2;
```
dshulyak, line 25: failure to be replaced by special events that have to be watched by clients

spacemesh/v1/admin_types.proto
dshulyak: decided to keep it to simplify integration, in future @brusherru may use better messages in smapp

spacemesh/v1/admin_types.proto
dshulyak: decided to keep failure to avoid keeping a list of events on the client side

spacemesh/v1/admin_types.proto
```diff
@@ -15,3 +17,84 @@ message RecoverRequest {
...
+  string help = 3;
+  oneof details {
+    EventBeacon beacon = 4;
+    EventInitStart init_start = 5;
+    EventInitComplete init_complete = 6;
+    EventPostStart post_start = 7;
+    EventPostComplete post_complete = 8;
+    EventPoetWait poet_wait = 9;
+    EventPoetWaitChallenge poet_wait_challenge = 10;
```
poszu, line 35: If I understand correctly, this is supposed to notify that nipostBuilder is waiting for poet round to finish, right? If that's the case, consider:
```suggestion
    EventPoetWaitForProof poet_wait_for_proof = 10;
```

spacemesh/v1/admin.proto
```diff
@@ -12,4 +12,7 @@ service AdminService {
 
   // Recovers from the provided checkpoint data.
   rpc Recover(RecoverRequest) returns (google.protobuf.Empty);
+
+  // Events that are relevant for node operator
+  rpc EventsStream(EventStreamRequest) returns (stream Event);
```
poszu, line 17: I thought there will be a separate stream of events for each in-node process, to not mix events coming from, say atxBuilder building process and beacon.

Will smapp show the record of all previous events? I imagine that, if the atxBuilder emitted `EventPoetWaitChallenge` then Smapp should show a big label "waiting for poet proof" for ~2 weeks regardless of any other event showing about after it.

spacemesh/v1/admin_types.proto
```diff
@@ -15,3 +17,84 @@ message RecoverRequest {
...
+    EventAtxPubished atx_published = 11;
```
poszu, line 36: How about adding an `EventPoetProofSelected/Received` to mark the point between receiving the proof and publishing the ATX?

poszu (approved):

spacemesh/v1/admin_types.proto
dshulyak: i can change if you prefer. challenge was in the context of the post, because what we receive from poet is used as a challenge in the post

spacemesh/v1/admin_types.proto
dshulyak: there will be PostStart/PostComplete event before AtxPublished, that says something "building post with challenge XXX". i think it would be sufficient to explain that it is done when poet is received

spacemesh/v1/admin.proto
dshulyak: > I thought there will be a separate stream of events for each in-node process, to not mix events coming from, say atxBuilder building process and beacon.

i can't just justify doing that. there is a limited number of events that are interesting for node admin. if we see that they grow to some huge number then certainly

> Will smapp show the record of all previous events? I imagine that, if the atxBuilder emitted EventPoetWaitChallenge then Smapp should show a big label "waiting for poet proof" for ~2 weeks regardless of any other event showing about after it.

i imagine that there will be a log events. like
- waiting for poet
- published proposal
- published proposal

so maybe waiting for poet won't be the last one. @brusherru please consider

spacemesh/v1/admin_types.proto
```diff
@@ -15,3 +17,84 @@ message RecoverRequest {
...
+
+message ProposalEligibility {
+  uint32 layer = 1;
+  uint32 count = 2;
+}
+
+// Published proposal. Rewards will be received, once proposal is included into the block.
+message EventProposal {
+  uint32 layer = 1;
+  bytes proposal = 2;
```
lrettig, line 97: do we not need to include weight or number of eligibilities here? to aid in calculating expected reward

spacemesh/v1/admin.proto
lrettig: what about the scenario where Smapp is opened and connects to an already-running go-sm node? it could be a long time in between events in the stream. don't we need a separate, non-stream endpoint to query the current/latest status?

spacemesh/v1/admin_types.proto
dshulyak: rewards are distributed based on relative weight in the block, i don't think that estimating makes any sense

spacemesh/v1/admin.proto
dshulyak: i realized that smesher will miss some events today, i wont be adding status but i will cache latest events

PR #259, opened 2023-08-16
add malfeasance proofs

countvonzero:


spacemesh/v1/mesh_types.proto
```diff
@@ -107,3 +107,21 @@ message EpochStreamRequest {
 message EpochStreamResponse {
     ActivationId id = 1;
 }
+
+message MalfeasanceRequest {
+    SmesherId smesher = 1;
+    string smesher_hex = 2;
```
fasmat, line 113: Are these both encoding the NodeID? Can we remove one of the two (otherwise what would happen if they are not the same and both valid?)

countvonzero: thanks. i think i will keep the hex version for UX.

fasmat: I would have gone with the other one for consistency 😅, but I'm OK with either as long as we only have one.

fasmat (approved):

spacemesh/v1/mesh_types.proto
countvonzero: so the machines (explorer/dashboard/smapp) will use the streaming API.
the query is mainly for human, hence the choice.

countvonzero:
@brusherru 
smapp can use the activation service to query ATX, which will have a malfeasance proof attached if present. 
or make query for malfeasance proof directly for the smsher id in hex.
there is also going to be node event when a node detect its own malfeasance proof.

PR #268, opened 2023-09-18
Add PoST Service API

### LINKED ISSUE #269: Draft API for Proof generation
### Description

The API needs to be extended to allow PoST services as described in https://github.com/spacemeshos/pm/issues/259 to connect to a node.

PoST services (clients) connect to the node (server), but the node sends requests to the services. This essentially means that requests are generally sent server -> client and responses client -> server.

### Acceptance criteria

- The gRPC API is extended with a new service for Post.
- This service needs to expose a bi-directional stream between the client and the server, where the server can issue multiple distinct requests and the client responds to those queries.
- The initial set of requests must include:
  - Request(s) to trigger proof generation for a given challenge and querying the proof after generation completes
  - Request(s) to query information / metadata about a service, including at least the following:
    - the node Id the PoST data was generated for
    - the size of the PoST data
    - the VRF Nonce

### Implementation hints

- For infos on bi-directional streams see https://grpc.io/docs/what-is-grpc/core-concepts#bidirectional-streaming-rpc

fasmat:
Closes #269


spacemesh/v1/post.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto3";
+package spacemesh.v1;
+option go_package = "github.com/spacemeshos/api/release/go/spacemesh/v1";
+
+import "spacemesh/v1/post_types.proto";
+
+// PoSTService is used by dedicated PoST nodes to interact with the spacemesh
+// node.
+service PoSTService {
```
poszu, line 9: The [style guide](https://protobuf.dev/programming-guides/style/#services) suggests using PascalCase. Using `PoST` leads to weird generated code like module name: `spacemesh_v1::po_st_service_client`.

```suggestion
// PostService is used by dedicated PoST nodes to interact with the spacemesh
// node.
service PostService {
```

spacemesh/v1/post_types.proto
```diff
@@ -0,0 +1,55 @@
+syntax = "proto3";
+package spacemesh.v1;
+option go_package = "github.com/spacemeshos/api/release/go/spacemesh/v1";
+
+import "spacemesh/v1/types.proto";
+
+// NodeRequest is a request sent from the node to the post service.
+message NodeRequest {
+  oneof type {
```
poszu, line 9: `type` is a keyword in many languages (both Go and Rust) and it forces the code gen to workaround this (in rust the field name becomes `r#type` :( How about using "kind":
```suggestion
// NodeRequest is a request sent from the node to the post service.
message NodeRequest {
  oneof kind {
```

spacemesh/v1/post.proto
```diff
@@ -0,0 +1,12 @@
...
+package spacemesh.v1;
+option go_package = "github.com/spacemeshos/api/release/go/spacemesh/v1";
+
+import "spacemesh/v1/post_types.proto";
+
+// PostService is used by dedicated PoST nodes to interact with the spacemesh
+// node.
+service PostService {
+  // Register allows a dedicated PoST node to connect to the spacemesh node.
+  rpc Register(stream ServiceResponse) returns (stream NodeRequest);
```
lrettig, line 11: This is confusing. Why does `Register` receive a "Response" and return a "Request"?

spacemesh/v1/post_types.proto
```diff
@@ -0,0 +1,55 @@
...
+
+message ProofMetadata {
+  bytes challenge = 1;
+  SmesherId node_id = 2;
+  ActivationId commitment_atx_id = 3;
+  uint32 num_units = 4;
+  uint64 labels_per_unit = 5;
+}
+
+enum GenProofStatus {
```
lrettig, line 45: What about proof generation in progress? Don't we want to be able to report the progress?

spacemesh/v1/post.proto
fasmat: Because the PoST Service connects as client to the node (server) but the node sends requests and receives responses from the service.

fasmat: This is a bi-directional stream: https://grpc.io/docs/what-is-grpc/core-concepts#bidirectional-streaming-rpc but in our case the node is always the one making requests and the post service is always the one responding.

spacemesh/v1/post_types.proto
fasmat: We are currently figuring this out as we go. That's why the PR is still a draft 🙂. The current idea is that the node is stateless and in regular intervals sends the same "GenerateProof" request to all PoST services connected. These requests are idempotent, so if a service receives a request with the same challenge twice it just responds OK and continues to generate the proof.

At the moment we also don't have any progress information about proof generation and we want to keep the change simple to be able to deliver earlier.

lrettig: I think it's important to be able to report the status of the proof generation while in progress, either via "push" (PoST service reports back with status periodically) or "pull" (node checks in with PoST service periodically). Proof generation takes hours and it's important the the user be able to see what's going on. See https://github.com/spacemeshos/go-spacemesh/issues/4742.

fasmat: Thats out of scope for this in my opinion. There is no easy way to know the current status of the proof generation.

The post service can keep track of how far into reading all data from disk it is, but this cannot be extrapolated to how long it will take for the proof generation to be finished. Because

- it takes x amount of time to read the first 50% of data; this does not mean that it takes the same amount of time to read the remaining data - depending on the disk and file system this could be quite different
- Just because it read 50% of the data doesn't mean the proof generation is done - it could take 1 pass or it could take 10 passes (depending on number of nonces per pass, etc.) to generate a proof. It's probabilistic there is no reliable way to tell how much more time it will take until a proof will be available.

We could report back to the node how much data has been read from disk already and which pass over the data it is, but how would we display that information to the user?

I would prefer to first discuss on how we measure progress, how we display it to the user and then design the interface around that and not the other way round 🙂

fasmat: Additionally the post service logs progress about proof generation via `os.Stderr` and if used in standalone mode that output will be captured and logged by the node as debug logs.

If you think this should be info logs instead I can change that 🙂

poszu: @fasmat 
>I would prefer to first discuss on how we measure progress, how we display it to the user and then design the interface around that and not the other way round 🙂

100% agree. Measuring POST proof generation progress is not trivial.

Note that this is not the final version of this API and we can iterate on it, adding more features as we go.

fasmat:
@poszu if you agree I would like to merge the current API as is. The initial implementation seems to work well and we can keep iterating on it in the future. With the API released we can merge https://github.com/spacemeshos/post-rs/pull/133, release that version and integrate it into the node.

wdyt?

spacemesh/v1/post.proto
fasmat: added clarifying comment.

poszu (approved):

PR #274, opened 2023-09-28
Update protobuf used for code generation

fasmat:
This updates the `Makefile` to use a more recent version of `buf` and `protoc` and generally simplifies it.

By adding `googleapis` to `buf.yaml` we don't need to keep copies of those in the repo as `buf` will fetch them for us during code generation.


Makefile
```diff
@@ -162,43 +113,14 @@ ssh: $(BUF)
...
-# clean deletes any files not checked in and the cache for all platforms.
-
-.PHONY: clean
-clean:
-	git clean -xdf
-	rm -rf $(CACHE_BASE)
+check:
+	@git diff --quiet || (echo "\033[0;31mWorking directory not clean!\033[0m" && git --no-pager diff && exit 1)
+	@make build
+	@git diff --name-only --diff-filter=AM --exit-code . || { echo "\nPlease rerun 'make generate' and commit changes.\n"; exit 1; }
```
poszu, line 126:
```suggestion
	@git diff --name-only --diff-filter=AM --exit-code . || { echo "\nPlease rerun 'make build' and commit changes.\n"; exit 1; }
```

release/go/go.mod
```diff
@@ -3,16 +3,16 @@ module github.com/spacemeshos/api/release/go
...
-	github.com/golang/protobuf v1.5.3
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.18.0
-	google.golang.org/genproto/googleapis/rpc v0.0.0-20230822172742-b8732ec3820d
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20230920204549-e6e6cdab5c13
 	google.golang.org/grpc v1.58.2
 	google.golang.org/protobuf v1.31.0
 )
 
 require (
+	github.com/golang/protobuf v1.5.3 // indirect
```
poszu, line 13: Where does it come from? I hoped that we could get rid of it finally :cry:

Makefile
```diff
@@ -24,112 +18,69 @@ HTTPS_GIT := https://github.com/spacemeshos/api.git
...
-CACHE_BIN := $(CACHE)/bin
-# Marker files are put into this directory to denote the current version of binaries that are installed.
-CACHE_VERSIONS := $(CACHE)/versions
-# We do some temporary work here
-CACHE_TMP := $(CACHE_BASE)/tmp
+install-buf:
+	@mkdir -p $(BIN_DIR)
+	curl -sSL "https://github.com/bufbuild/buf/releases/download/v$(BUF_VERSION)/buf-$(UNAME_OS)-$(UNAME_ARCH)" -o $(BIN_DIR)/buf
+	@chmod +x $(BIN_DIR)/buf
+.PHONY: install-buf
```
poszu, line 61: The existing approach with "cache" has a nice property that it automatically detects if a new version needs to be installed. That's lost now :(

Although it could be ~easily restored. How about:
```suggestion
BUF := $(BIN_DIR)/buf_$(BUF_VERSION)
$(BUF):
	@mkdir -p $(BIN_DIR)
	curl -sSL "https://github.com/bufbuild/buf/releases/download/v$(BUF_VERSION)/buf-$(UNAME_OS)-$(UNAME_ARCH)" -o $@
	@chmod +x $@
```

And then use it like that:

```makefile
build: $(BUF)
	$(BUF) generate
```

Now, buf will be automatically updated (or downloaded if doesn't exist) on `make build` . No need for `make install` at all.

:bulb: It could also remove existing bufs before downloading the new version:
```makefile
	rm $(BIN_DIR)/buf_*
```

poszu (commented):
:+1: 

How about using buf's linting features too?

release/go/go.mod
fasmat: The reason is that `google.golang.org/grpc` and `github.com/grpc-ecosystem/grpc-gateway/v2` depend on it. 

I assume to ensure compatibility between those libraries and `github.com/golang/protobuf` until the later is deprecated.

```
$ go mod graph | grep github.com/golang/protobuf@v1.5.3
...
github.com/grpc-ecosystem/grpc-gateway/v2@v2.18.0 github.com/golang/protobuf@v1.5.3
google.golang.org/genproto/googleapis/api@v0.0.0-20230920204549-e6e6cdab5c13 github.com/golang/protobuf@v1.5.3
google.golang.org/grpc@v1.58.2 github.com/golang/protobuf@v1.5.3
```

Makefile
fasmat: I wouldn't overcomplicate it. We can do a similar approach as we do with our own dependencies: have the zip file containing buf as a `make` target that downloads and extracts the executable. If the zip file is already there it will be a no-op and otherwise the executable will be updated, wdyt?

fasmat:
> 👍
> 
> How about using buf's linting features too?

We already do, the following `make` targets all call `buf lint`:

- `make https`
- `make ssh`
- `make local`

spacemesh/v1/api_config.yaml
```diff
@@ -63,7 +63,7 @@ http:
     - selector: spacemesh.v1.NodeService.Status
       post: "/v1/node/status"
```
fasmat, line 64: Move these selectors into the proto files

Makefile
fasmat: I did something similar: `buf` is now installed with version appended to `bin/buf_*` and the file is a target of the `Makefile`. Updating the `buf` version will download the new version and delete the previous.

Additionally `protoc` isn't installed any more, it's not actually needed (only its plugins).

fasmat:
`make install` has been removed (not needed any more), as well as `make ssh` (no need to access repo via ssh if https is possible.

`make https` has been renamed to `make breaking`.

`make lint` is now running the linter (not done again when calling `make breaking`).

I updated the README.md as well.

poszu (approved):

PR #291, opened 2023-12-19
fix protobuf generate to java classes got conflicting

liyifeng:
fix protobuf generate to java classes got conflicting


liyifeng:
same to #284

Issue #295, opened 2024-01-02
Prepare API v2 specification


kacpersaw:

brusherru:
Here are the list of APIs required by Wallet application:

**Network-specific data**
- Node info: `node/nodeinfo`
  To get a proper HRP,
  First genesis layer & epoch size for timestamps
- GenesisTime: `mesh/genesistime`
  To calculate timestamps
- LayerDuration: `mesh/layerduration`
  To calculate timestamps

**Node status:**
- Node status: `node/status`
  To ensure that Node is synced & can publish txs

**User data:**
- Balance: `globalstate/accountdataquery` (can be replaced with `globalstate/account`)
- Txs:
  - Tx data: `mesh/accountmeshdataquery`
  - Tx states: `transaction/transactionsstate`
- Rewards: `globalstate/accountdataquery`

**Publishing txs:**
- Gas estimation: `transaction/parsetransaction`
   _Funny thing that it may parse TX without a signature, but then it will estimate the wrong gas amount, like it may be published without a signature. So if you don't have a signature yet — put there 64 zero-bytes and `verify: false` :)
   This statement is valid for gRPC as well._
- Submit: `transaction/submittransaction`

All of them already exist and should work via JSON RPC (I haven't tested all of them yet).
So I'm using them now, and propose to focus on endpoints that consume too much resources or are unsafe.

kacpersaw:
### Done
* Activation service: https://github.com/spacemeshos/api/pull/305
* Reward service: https://github.com/spacemeshos/api/pull/307
* Network service: https://github.com/spacemeshos/api/pull/315

### In progress:
* Node service: https://github.com/spacemeshos/api/pull/316
* Layer and block service: https://github.com/spacemeshos/api/pull/319

### Pending
* Account service
* Layer and block service
* Account service
* Transaction service

kacpersaw:
API is implemented and exposed on https://mainnet-api-docs.spacemesh.network/index.html

PR #300, opened 2024-01-24
API v2 specification along with the new requirements

kacpersaw:


spacemesh/v2/node.proto
```diff
@@ -0,0 +1,18 @@
+syntax = "proto3";
+
+package spacemesh.v2;
+
+message NodeStatusRequest {}
+
+message NodeStatusResponse {
+  uint64 connected_peers = 1; // number of connected neighbors
+  bool   is_synced = 2; // true when meshed is synced
```
lrettig, line 9: I think we need one or two additional modes here:

- status unknown
- offline
- syncing
- synced

spacemesh/v2/node.proto
```diff
@@ -0,0 +1,18 @@
...
+  uint32 synced_layer = 3; // the last layer node has synced
+  uint32 verified_layer = 4; // the last layer node has verified
+  uint32 head_layer = 5; // head layer is the tip
```
lrettig, line 12: how is this different than verified_layer?

spacemesh/v2/node.proto
```diff
@@ -0,0 +1,18 @@
...
+  uint32 head_block = 6;
```
lrettig, line 13: I think we need to add something like `current_layer` which is the current layer number based strictly on time

spacemesh/v2/reward.proto
```diff
@@ -0,0 +1,32 @@
...
+  oneof versioned {
+    RewardV1 v1 = 1;
+  }
+}
+
+message RewardV1 {
+  uint32 layer = 1; // layer award was paid in
+  uint64 total = 2; // total reward paid (sum of tx fee and layer reward)
+  uint64 layer_reward = 3; // tx_fee = total - layer_reward
+  uint32 layer_computed = 4; // layer number of the layer when reward was computed
```
lrettig, line 15: drop this for now, we don't use this

spacemesh/v2/reward.proto
```diff
@@ -0,0 +1,32 @@
...
+  string coinbase = 5; // account awarded this reward
+  bytes smesher = 6;  // id of smesher who earned this reward
```
lrettig, line 17: bytes or string format?

spacemesh/v2/reward.proto
```diff
@@ -0,0 +1,32 @@
...
+
+message Reward {
+  oneof versioned {
+    RewardV1 v1 = 1;
+  }
+}
+
+message RewardV1 {
+  uint32 layer = 1; // layer award was paid in
+  uint64 total = 2; // total reward paid (sum of tx fee and layer reward)
```
lrettig, line 13:
```suggestion
  uint64 total = 2; // total reward paid in smidge (sum of tx fee and layer reward)
```

spacemesh/v2/reward.proto
```diff
@@ -0,0 +1,32 @@
...
+  uint64 layer_reward = 3; // tx_fee = total - layer_reward
+  uint32 layer_computed = 4; // layer number of the layer when reward was computed
+  string coinbase = 5; // account awarded this reward
+  bytes smesher = 6;  // id of smesher who earned this reward
+}
+
+message RewardRequest {
```
lrettig, line 20: no way to query rewards by smesherID?

spacemesh/v2/reward.proto
```diff
@@ -0,0 +1,32 @@
...
+  string coinbase = 1;
+  uint64 offset = 2;
```
lrettig, line 22: can we document `offset` and `limit` in comments here?

spacemesh/v2/tx.proto
```diff
@@ -0,0 +1,114 @@
...
+  oneof versioned {
+    TransactionV1 v1 = 1;
+  }
+}
+
+// An immutable Spacemesh transaction.
+// do not include mutable data such as tx state or result.
+message TransactionV1 {
+  bytes id = 1;
+  string principal = 2;
```
lrettig, line 17:
```suggestion
  string principal = 2; // principal account address
```

spacemesh/v2/tx.proto
```diff
@@ -0,0 +1,114 @@
...
+  string template = 3;
```
lrettig, line 18:
```suggestion
  string template = 3; // template account address
```

spacemesh/v2/tx.proto
```diff
@@ -0,0 +1,114 @@
...
+  uint32 method = 4; // this is actually limited by uint8, but no type for that.
+  Nonce nonce = 5;
+  LayerLimits limits = 6;
```
lrettig, line 21: need to double check but pretty sure this is unused, not sure we want to include this

spacemesh/v2/tx.proto
```diff
@@ -0,0 +1,114 @@
...
+  uint64 max_gas = 7;
+  uint64 gas_price = 8;
```
lrettig, line 23:
```suggestion
  uint64 gas_price = 8; // fee per unit gas, in smidge
```

spacemesh/v2/tx.proto
```diff
@@ -0,0 +1,114 @@
...
+// An immutable Spacemesh transaction.
+// do not include mutable data such as tx state or result.
+message TransactionV1 {
+  bytes id = 1;
+  string principal = 2;
+  string template = 3;
+  uint32 method = 4; // this is actually limited by uint8, but no type for that.
+  Nonce nonce = 5;
+  LayerLimits limits = 6;
+  uint64 max_gas = 7;
```
lrettig, line 22:
```suggestion
  uint64 max_gas = 7; // max gas units consumed by tx
```

spacemesh/v2/tx.proto
```diff
@@ -0,0 +1,114 @@
...
+message ParseTransactionRequest {
+  bytes transaction = 1;  // signed binary transaction
+  bool verify = 2;        // if true signature verification will be executed
+}
+
+message SubmitTransactionRequest {
+  bytes  transaction = 1;  // signed binary transaction
+}
+
+message SubmitTransactionResponse {
```
lrettig, line 105: this should return failure, or else a txid. I think that's it. need to check but assuming handling of incoming txs is async (which I think it is), then it can't return tx state.

spacemesh/v2/tx.proto
```diff
@@ -0,0 +1,114 @@
...
+  google.rpc.Status status = 1;
+  TransactionState tx_state = 2;
+}
+
+service TransactionService {
+  rpc Get(TransactionRequest) returns (TransactionResponse);
+  rpc ParseTransaction(ParseTransactionRequest) returns (ParseTransactionRequest);
```
lrettig, line 112: need a `ParseTransactionResponse` here - which I guess should return a tx object (or else an error if it's invalid/malformed)

spacemesh/v2/tx.proto
```diff
@@ -0,0 +1,114 @@
...
+  LayerLimits limits = 6;
+  uint64 max_gas = 7;
+  uint64 gas_price = 8;
+  uint64 max_spend = 9;
+  bytes raw = 10;
+}
+
+message Nonce {
+  uint64 counter = 1;
+  uint32 bitfield = 2; // this is actually limited by uint8, but no type for that.
```
lrettig, line 30: need to check whether this is actually used

spacemesh/v2/tx.proto
```diff
@@ -0,0 +1,114 @@
...
+  enum Status {
+    SUCCESS = 0;
+    FAILURE = 1;
+    INVALID = 2;
+  }
+
+  Transaction tx = 1;
+  Status status = 2;
+  string message = 3;
+  uint64 gas_consumed = 4;
```
lrettig, line 65:
```suggestion
  uint64 gas_consumed = 4; // in units of gas
```

spacemesh/v2/tx.proto
```diff
@@ -0,0 +1,114 @@
...
+  uint64 fee = 5;
```
lrettig, line 66:
```suggestion
  uint64 fee = 5; // in smidge
```

spacemesh/v2/tx.proto
```diff
@@ -0,0 +1,114 @@
...
+  bytes block = 6;
+  uint32 layer = 7;
+  repeated string touched_addresses = 8;
```
lrettig, line 69: should this return `Account` objects?

spacemesh/v2/tx.proto
```diff
@@ -0,0 +1,114 @@
...
+
+message Transaction {
+  oneof versioned {
+    TransactionV1 v1 = 1;
+  }
+}
+
+// An immutable Spacemesh transaction.
+// do not include mutable data such as tx state or result.
+message TransactionV1 {
```
lrettig, line 15: this needs to include a status field:

```
  enum Status {
    UNKNOWN = 0; // block hasn't been processed yet
    SUCCESS = 1; // tx was fully applied
    FAILURE = 2; // tx failed, but base gas was paid
    INEFFECTIVE = 3; // tx wasn't processed, no gas paid
  }
```

lrettig, line 15: need to also add recipient and send amount (for Send txs). I'm not 100% sure how to structure this. we should probably add a subfield, something liked ParsedContents, that's `oneof` depending on the type of tx (spawn, self spawn, send, etc.)

lrettig: actually, I think we can just include the `TransactionResult` here, which includes these things.

lrettig (changes requested):
great start. see comments.

one more thing to add: `TransactionService.EstimateGas` which receives a tx (with max gas unset) and returns a recommended max gas.

spacemesh/v2/block.proto
```diff
@@ -0,0 +1,26 @@
...
+
+message Block {
+  oneof versioned {
+    BlockV1 v1 = 1;
+  }
+}
+
+message BlockV1 {
+  bytes id = 1; // block hash
+  repeated Transaction transactions = 2; // block transactions
```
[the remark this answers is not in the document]
kacpersaw: the order of records for the same field with respect to each other is preserved
 https://protobuf.dev/programming-guides/encoding/#optional

spacemesh/v2/account.proto
```diff
@@ -0,0 +1,28 @@
...
+  AccountState state_projected = 3; // projected state (includes pending txs)
+}
+
+message AccountState {
+  uint64 counter = 1; // aka nonce
+  uint64 balance = 2; // known account balance
+}
+
+message AccountRequest {
+  string address = 1;
```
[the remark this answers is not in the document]
kacpersaw: I think it will be more user friendly to use string, and we already use string in V1. https://github.com/spacemeshos/api/blob/master/spacemesh/v1/types.proto#L19

spacemesh/v2/layer.proto
```diff
@@ -0,0 +1,35 @@
...
+
+package spacemesh.v2;
+
+message Layer {
+  oneof versioned {
+    BlockV1 v1 = 1;
+  }
+}
+
+message LayerV1 {
```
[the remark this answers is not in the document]
kacpersaw: What about layer_hash and root_state_hash?

spacemesh/v2/network.proto
```diff
@@ -0,0 +1,20 @@
...
+
+message NetworkInfoRequest {}
+
+message NetworkInfoResponse {
+  uint64 genesis_time = 1; // network genesis time as unix epoch time
+  uint32 epoch_num_layers = 2; // number of layers per epoch
+  uint64 layer_duration = 3; // layer duration, in seconds
+  uint64 max_transactions_per_second = 4;
+  string hrp = 5;
+  uint32 first_genesis = 6;
```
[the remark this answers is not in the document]
kacpersaw: I'm not sure, it was used here https://github.com/spacemeshos/api/blob/master/spacemesh/v1/node_types.proto#L70

spacemesh/v2/network.proto
```diff
@@ -0,0 +1,20 @@
...
+  uint32 effective_genesis = 7;
+  uint32 epoch_size = 8;
```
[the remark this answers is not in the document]
kacpersaw: https://github.com/spacemeshos/api/blob/master/spacemesh/v1/node_types.proto#L72

spacemesh/v2/reward.proto
kacpersaw: As we use string for address, I will change that to string as well

spacemesh/v2/tx.proto
kacpersaw: I want to separate transaction from mutable data like tx state and result.
That's why I've created wrapper around it. https://github.com/spacemeshos/api/blob/api-v2/spacemesh/v2/tx.proto#L90

spacemesh/v2/layer.proto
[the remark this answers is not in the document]
kacpersaw: What about `hash` and `root_state_hash`?

spacemesh/v2/network.proto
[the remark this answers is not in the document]
kacpersaw: `first_genesis` returns the first effective genesis layer

spacemesh/v2/network.proto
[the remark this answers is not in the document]
kacpersaw: `epoch_size` returns number of layers per epoch

spacemesh/v2/node.proto
kacpersaw: `verified_layer` is the latest verified layer by the node
`head_layer` will be the top layer, based on time

spacemesh/v2/node.proto
kacpersaw: in v1 we have

`synced_layer` - latest layer we saw from the network
`top_layer` - current layer, based on time
`verified_layer` - latest verified layer

what is your proposal?

spacemesh/v2/tx.proto
kacpersaw: I think it's not used currently
https://github.com/spacemeshos/go-spacemesh/blob/416a313709be5eb41d2c2bf3d6a90605e30208c3/api/grpcserver/mesh_service.go#L239-L239

spacemesh/v2/block.proto
```diff
@@ -0,0 +1,26 @@
...
+
+package spacemesh.v2;
+
+message Block {
+  oneof versioned {
+    BlockV1 v1 = 1;
+  }
+}
+
+message BlockV1 {
```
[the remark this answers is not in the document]
kacpersaw: In v1, full transactions are sent within a block

spacemesh/v2/tx.proto
kacpersaw: we can also left that to user

spacemesh/v2/reward.proto
kacpersaw: added second handler to list by smesher. Probably we can combine this into one.

spacemesh/v2/network.proto
```diff
@@ -0,0 +1,20 @@
...
+
+message NetworkInfoRequest {}
+
+message NetworkInfoResponse {
+  uint64 genesis_time = 1; // network genesis time as unix epoch time
+  uint32 epoch_num_layers = 2; // number of layers per epoch
+  uint32 layer_duration = 3; // layer duration, in seconds
+  bytes genesis_id = 4;
+  string hrp = 5;
+  uint32 first_genesis = 7; // first effective genesis layer
```
lrettig, line 13: I see in the code that this has to do with checkpoint restore. In practice effective genesis and "first genesis" are the same. Let's keep things simple and collapse these two into a single data item called `effective_genesis_layer` for clarity.

spacemesh/v2/network.proto
```diff
@@ -0,0 +1,20 @@
...
+  uint32 effective_genesis = 8; // last layer of genesis
```
lrettig, line 14:
```suggestion
  uint32 effective_genesis_layer = 8; // effective genesis layer, i.e., first layer after genesis initialization period
```

spacemesh/v2/network.proto
```diff
@@ -0,0 +1,20 @@
...
+  uint32 epoch_size = 9; // number of layers per epoch
```
lrettig, line 15:
```suggestion
  uint32 layers_per_epoch = 9; // number of layers per epoch
```

spacemesh/v2/node.proto
lrettig: I think we're missing one. Here's my proposal:

```
  uint32 synced_layer = 3; // last layer node has synced
  uint32 verified_layer = 4; // last layer node has verified
  uint32 head_layer = 5; // current chain head; the last layer the node has gossiped or seen
  uint32 current_layer = 6; // current layer, based on clock time
```

`synced_layer` is only useful while syncing. Blocks are also received via gossip. The relationship should be:

synced_layer <= verified_layer <= head_layer <= current_layer

spacemesh/v2/tx.proto
lrettig: Yes, this is an important requirement for consumers. We shouldn't require them to parse a raw TX, the API should provide the parsed TX. If for some reason we think this will be expensive, we can add a flag to the Request object to enable this.

spacemesh/v2/tx.proto
lrettig: confirmed, layer limits for tx are not currently enabled. let's remove this for now for simplicity and to avoid confusion.

lrettig (changes requested):
Great work, design is almost done.

lrettig:
See latest commit, this is my attempt to add parsed tx contents. Feel free to tweak!

kacpersaw:
> See latest commit, this is my attempt to add parsed tx contents. Feel free to tweak!

looks good, we can start with that

spacemesh/v2alpha1/node.proto
```diff
@@ -13,10 +13,11 @@ message NodeStatusResponse {
...
   SyncStatus status = 2;
-  uint32 synced_layer = 3; // the last layer node has synced
-  uint32 verified_layer = 4; // the last layer node has verified
-  uint32 head_layer = 5; // head layer is the tip
-  uint32 head_block = 6;
+  uint32 synced_layer = 3; // last layer node has synced
+  uint32 verified_layer = 4; // last layer node has verified
+  uint32 head_layer = 5; // current chain head; the last layer the node has gossiped or seen
+  uint32 current_layer = 6; // current layer, based on clock time
+  uint32 head_block = 7;
```
lrettig, line 20: Blocks don't have numeric IDs. Let's rename it `head_block_id` and make it either a data blob (binary) or else a string.

spacemesh/v2alpha1/tx.proto
```diff
@@ -2,14 +2,36 @@ syntax = "proto3";
...
+message ContentsSingleSigSpawn {
+  string pubkey = 1;
+}
+
+message ContentsMultiSigSpawn {
+  uint32 required = 1; // number of required signatures
+  repeated string pubkey = 2;
+}
+
+message TransactionContents {
```
lrettig, line 27: This should also be versioned, but it lives inside a versioned tx type so maybe that's enough. What do you think?

spacemesh/v2alpha1/tx.proto
```diff
@@ -18,22 +40,27 @@ message TransactionV1 {
...
-  LayerLimits limits = 6;
-  uint64 max_gas = 7; // max gas units consumed by tx
-  uint64 gas_price = 8; // fee per unit gas, in smidge
-  uint64 max_spend = 9;
-  bytes raw = 10;
+  uint64 max_gas = 6; // max gas units consumed by tx
+  uint64 gas_price = 7; // fee per unit gas, in smidge
+  uint64 max_spend = 8;
+  bytes raw = 9;
+  enum TransactionType {
```
lrettig, line 47: For the record there are other types, involving the vault and vesting wallet templates, but neither is relevant right now and we can add these later.

lrettig (changes requested):
lgtm! just need one more small change.

kacpersaw:
Specification was divided into separate PRs to avoid merging all definitions at once. Progress is noted here: https://github.com/spacemeshos/api/issues/295#issuecomment-2032459220

PR #302, opened 2024-01-31
Extend API for multismesher node

fasmat:
For multi-smeshing support the API has to be updated such that the node can report events correctly to clients about which Identity started / finished post and which Identities are managed by the node.

I updated all types and endpoints in a backwards compatible way and marked outdated APIs as depreciated.

cc @brusherru


brusherru:
"Multi-smeshing" is the same as running a node with multiple post services?

As for Node events, it is clear that I can just display a smesher id.

But I'm not sure should Smapp display more than one smesher id on the smesher screen 🤔 
Or I can just keep using deprecated API in Smapp, and this new feature will be used in a newer Node Manager app 🤔 
What do you think?

brusherru (approved):
LGTM

fasmat:
> "Multi-smeshing" is the same as running a node with multiple post services?

Yes

> As for Node events, it is clear that I can just display a smesher id.
>
> But I'm not sure should Smapp display more than one smesher id on the smesher screen 🤔
Or I can just keep using deprecated API in Smapp, and this new feature will be used in a newer Node Manager app 🤔
What do you think?

The deprecated API will always return 00..00 as smesher ID. In the case where the node runs in supervised mode (what we are currently doing) the new API will return an array with only one SmesherID, but as soon as multiple post services are connected to the node it will return an array with all those IDs.

That's why I deprecated the old call, it doesn't make sense in this case to just return one - which one?

brusherru:
Okay, that makes sense.

I just thought that Smapp also displays some other stuff (for example a size of PoST, location, etc) and when we have a list of Smesher IDs it will be not clear which one corresponds to this data...
And I see that there are some options:
- a). do not display the rest data if there is more than one SmehserID
- b). have an additional API endpoint which will return details per SmesherID, e.g.
   ```
   SmesherDataRequest(SmesherID) -> SmesherDataResponse {
     dir: string;
     num_units: u64;
     ...
    }
    ```

But anyway, as for now — I will just display a list of Smesher IDs ;)

fasmat:
Right now the data you are referring to is provided via the `PostSetupStatus` endpoint. In the supervised setup (i.e. only one post service that is managed by the node itself) it will continue to work as it did before.

In a multi-smeshing setup that endpoint won't return anything as the node does not have this information. Instead the post services could be queried for that info

cc @poszu

brusherru:
Okay, got it

spacemesh/v1/smesher.proto
```diff
@@ -39,6 +39,14 @@ service SmesherService {
         option (google.api.http) = {
             post: "/v1/smesher/smesherid"
         };
+        option deprecated = true; // will return empty response in v1.4.x
```
poszu, line 42: Wouldn't it generate a compiler warning when you try to implement `SmesherID` service in go-sm?

fasmat: Afaik only on the client side - we do implement the deprecated handler (otherwise we wouldn't satisfy an interface) and the linter doesn't complain.

poszu:
@mafa
> The deprecated API will always return 00..00 as smesher ID. 

I think it would be better to return an "Unimplemented" status (perhaps with a message informing to switch to the new API) instead. It would become obvious that this API is no longer available and the user should switch to the new one. Otherwise, it might get overlooked and various tools will show `00000...` without their authors realizing that this API is no longer working as expected.

fasmat:
@poszu I changed the handler to return `code.Unimplemented` when calling the deprecated function.

poszu (approved):

PR #304, opened 2024-02-12
v2alpha1 stream for atxs

kacpersaw:
Based on #281 

Changed version from v2 to v2alpha1 and removed headers


spacemesh/v2alpha1/activation.proto
```diff
@@ -0,0 +1,93 @@
...
+}
+
+message PoetProof {
+    repeated bytes proof_nodes = 1;
+    uint64 leaf = 2;
+}
+
+message PostMeta {
+    bytes challenge = 1;
+    uint64 labels = 2;
```
poszu, line 42: Is it the `"labelsPerUnit"`? If so, I think it's not required here as this is a network parameter (all ATXs on the network have the same value). I know it is part of the Activation structure (it probably shouldn't be) so I guess its fine to keep it here for symmetry :shrug: .

spacemesh/v2alpha1/activation.proto
```diff
@@ -0,0 +1,93 @@
...
+    Post post = 15;
+    PostMeta post_meta = 16;
+    VRFPostIndex vrf_post_index = 17;
+}
+
+message VRFPostIndex {
+    uint64 nonce = 1;
+}
+
+message PoetProof {
```
poszu, line 35: NIT: Maybe rename it to `MembershipProof` or `PoetMembershipProof` to avoid confusion with the Poet proof.

PR #305, opened 2024-02-14
Add v2alpha1 atxs

kacpersaw:
Based on https://github.com/spacemeshos/api/pull/281

Changed version from v2 to v2alpha1 and removed headers


spacemesh/v2alpha1/activation.proto
```diff
@@ -0,0 +1,89 @@
...
+  }
+}
+
+message ActivationV1 {
+  bytes id = 1;
+  bytes node_id = 2;
+  bytes signature = 3;
+  uint32 publish_epoch = 4;
+  uint64 sequence = 5;
+  bytes prev_atx = 6;
```
poszu, line 17: NIT: consider naming it with full name to be consistent with other atxs below.
```suggestion
  bytes previous_atx = 6;
```

spacemesh/v2alpha1/activation.proto
```diff
@@ -0,0 +1,89 @@
...
+}
+
+message Post {
+  uint32 nonce = 1;
+  bytes indices = 2;
+  uint64 pow = 3;
+}
+
+message ActivationStreamRequest {
+  uint32 start_epoch = 1; // Apply EpochFilter together with CoinbaseFilter for better performance.
```
poszu, line 52: I still don't understand the part about EpochFilter and CoinbaseFilter. How to "apply epoch filter"?

poszu (approved):

PR #313, opened 2024-03-11
Add ChangeLogLevel to DebugService

kacpersaw:
Part of https://github.com/spacemeshos/go-spacemesh/issues/5674


dshulyak (approved):
api looks good. can we document here that "*" means all. 
we can consider returning all registered modules, but i think it is useful mostly for developers, and they should be able to get them from code

PR #314, opened 2024-03-13
Add Smesher ID to events

fasmat:
This adds the Smesher ID to events that are emitted by every identity managed by the node to be able to distinguish them from each other.


dshulyak (approved):

fasmat:
cc @brusherru the addition of the NodeID should be backwards compatible, but you can now display which identity triggered the event (since there will be multiple in a remote smeshing setup now).

PR #315, opened 2024-03-15
v2alpha1: add network service

kacpersaw:
API v2 spec https://github.com/spacemeshos/api/pull/300

Add new network service for v2alpha1 based on API v2 spec.


spacemesh/v2alpha1/network.proto
```diff
@@ -0,0 +1,18 @@
...
+  uint64 genesis_time = 1; // network genesis time as unix epoch time
+  uint32 layer_duration = 3; // layer duration, in seconds
+  bytes genesis_id = 4;
+  string hrp = 5;
+  uint32 effective_genesis_layer = 6; // effective genesis layer, i.e., first layer after genesis initialization period
+  uint32 layers_per_epoch = 7; // number of layers per epoch
+}
+
+service NetworkService {
+  rpc Info(google.protobuf.Empty) returns (NetworkInfoResponse);
```
poszu, line 17: How about using an empty `message NetworkInfoRequest {}` type instead? The advantage is that it could be later extended without breaking compatibility (in protobuf you can add new fields, but cannot change the type). IDK if it ever makes sense to pass anything in this request but it doesn't harm to be prepared.

spacemesh/v2alpha1/network.proto
```diff
@@ -0,0 +1,18 @@
+syntax = "proto3";
+
+import "google/protobuf/empty.proto";
+
+package spacemesh.v2alpha1;
+
+message NetworkInfoResponse {
+  uint64 genesis_time = 1; // network genesis time as unix epoch time
```
poszu, line 8: Why not use the `Timestamp` from the well-known-types?

spacemesh/v2alpha1/network.proto
```diff
@@ -0,0 +1,18 @@
...
+  uint32 layer_duration = 3; // layer duration, in seconds
```
poszu, line 9: Here you could use `Duration` from the WKT.

poszu (approved):

Issue #287, opened 2023-12-07
Block validity not currently exposed


lrettig:
In general there should be zero or one block per layer. On rare occasions there may be more than one, see, e.g., https://github.com/spacemeshos/go-spacemesh/issues/4825.

In these situations the API must make clear which block is the canonical block. go-spacemesh knows which one is canonical (https://github.com/spacemeshos/explorer-backend/issues/92#issuecomment-1841293498) but this is not currently exposed via the API.

Avionic23:
Hey, @lrettig. Sorry for bothering you again) I have a question related to block validity in case of multiple blocks in the layer. I've been reading documentation and noticed that only transactions from the valid block are processed by STF. Can we verify block validity by fetching transaction states from all layer blocks with '/spacemesh.v1.TransactionService/TransactionsState' RPC call and including only ones with  'TransactionState_TRANSACTION_STATE_PROCESSED' state? If yes, then how many layers should we wait before we can verify that?

lrettig:
@Avionic23 The way things work today with API v1, the API _only_ returns the canonical block for a given layer. See:

https://github.com/spacemeshos/go-spacemesh/blob/3302f7cf53c9364bda28ee6ac5a8b8e3d988a4bd/api/grpcserver/mesh_service.go#L277-L278

@kacpersaw we plan to do the same thing for API v2, correct? I'm referring to this comment: https://github.com/spacemeshos/api/pull/300#discussion_r1465528221.

At this time we don't have a plan for the API to return all blocks, including non-canonical ones. I don't see a use case, @Avionic23 do you?

Avionic23:
@lrettig @kacpersaw Yeah, I see that only one canonical block is always returned for the layer. What about transactions validity? I read in the documentation that there are possibilities of failed transactions in global state included in layer/block and I'm not sure how to figure them out. I see that it doesn't make sense to verify transactions with '/spacemesh.v1.TransactionService/TransactionsState' call because the 'LayersQuery' response already includes 'types.APPLIED' txs and this status is interchangeable with 'TransactionState_TRANSACTION_STATE_PROCESSED'. Please, tell me if I'm wrong about this and all txs with 'types.APPLIED' state are assumed to be valid

lrettig:
For API v1 please see https://github.com/spacemeshos/wiki/wiki/Integration-guide#tracking.

[...]

Issue #38, opened 2020-06-02
Handling reorgs


lrettig:
Up to now, @avive and I have been operating under the assumption that we don't need any special handling of reorgs. The various streams can just re-send all updated data and let the client sort it out.

However, we need to think this out a little more thoroughly as there are likely some corner cases we're not currently accounting for. It's a little easier to reason about the mesh, but what happens wrt global state? What if, e.g., an account with a balance disappears after a reorg? Etc.

@talm proposed that a stream sends a special "reorg" token, then dies, as an indication of a reorg. @noamnelke has some ideas for how to handle reorgs that he will share.

avive:
This might be a good way to this - when a 'reorg' event is sent, explorer-like clients should likely try to sync from genesis to get old blocks and transactions possible new state and wallet-like clients should query the api for fresh data for all the entities they store and cache locally. e.g. transactions and rewards.

noamnelke:
Not sure re-syncing from genesis is required (it might become very expensive after a while).

For immutable objects (transactions, ATXs, etc.) we just need to tag them with the layer in which they were added to the database. When a re-org happens we should be able to know the depth of the re-org (how many layers back it affects) and then invalidate all objects created after this layer and recreate them as we re-sync from that layer to the current layer.

For mutable objects, like accounts (balance, nonce, etc.) and contracts, we must be able to recreate the updated state. Intuitively, it seems possible to me, but I need to look at the exact data we collect to know for sure.

lrettig:
> For immutable objects (transactions, ATXs, etc.) we just need to tag them with the layer in which they were added to the database. When a re-org happens we should be able to know the depth of the re-org (how many layers back it affects) and then invalidate all objects created after this layer and recreate them as we re-sync from that layer to the current layer.

With respect to the API, though, and especially streams, this would require sending some sort of special "token" on the stream to indicate that a reorg happened, and the depth of the reorg. We then have two options:

- restream all updated data since the point of the reorg: this has the downside that it may overwhelm downstream clients that are slow to consume the data
- instead, continue to stream only new data, and expect downstream clients to use a historical "query" endpoint to resync all data since the point of the reorg

noamnelke:
Streaming only new data would be considerably harder to implement. The first option is what would happen "automatically".

[...]

PR #319, opened 2024-03-20
v2alpha1: Add LayerService with transaction and block definition

kacpersaw:
API v2 spec https://github.com/spacemeshos/api/pull/300

Add new layer service for v2alpha1 based on API v2 spec.


spacemesh/v2alpha1/layer.proto
```diff
@@ -0,0 +1,34 @@
...
+  uint32 number = 1; // layer number - not hash - layer content may change
+  enum LayerStatus {
+    LAYER_STATUS_UNSPECIFIED = 0; // not yet approved or confirmed
+    LAYER_STATUS_APPROVED = 1; // approved by hare
+    LAYER_STATUS_CONFIRMED = 2; // confirmed by tortoise
+  }
+  LayerStatus status = 2;
+  bytes hash = 3; // computer layer hash - do we need this?
+  Block blocks = 4; // layer's block
+  bytes root_state_hash = 5; // the root state hash of global state in this layer. if unset block was not applied optimistically.
```
dshulyak, line 23: i think this needs to be improved, i suggest following:
- consensus hash - for user to understand that they are on the right side of the fork (if such exists)
- state hash - fingeprint of the computed state at the layer
- cumulative state hash - cumulative fingerpint that uniquely identifies state since genesis
- block hash

layer should be reported when it was applied to the state.  i suggest to remove this unknown state. there should be a field to express that it was only applied because of hare, and not yet verified by tortoise . maybe `validity` which is equal to hare or tortoise

spacemesh/v2alpha1/layer.proto
```diff
@@ -0,0 +1,34 @@
...
+}
+
+message LayerRequest {
+  uint32 layer_number = 1;
+  bool include_transactions = 2;
```
dshulyak, line 28: i think they should be never included. report block, and provide list api to fetch transactions that are a part of the block in layer X

spacemesh/v2alpha1/layer.proto
```diff
@@ -0,0 +1,34 @@
...
+};
+
+
+service LayerService {
+  rpc Get(LayerRequest) returns (Layer);
```
dshulyak, line 33: i suggest to add proper List/Stream api like you did for several other services

spacemesh/v2alpha1/block.proto
```diff
@@ -0,0 +1,16 @@
...
+message Block {
+  oneof versioned {
+    BlockV1 v1 = 1;
+  }
+}
+
+message BlockV1 {
+  bytes id = 1; // block hash
+  repeated Transaction transactions = 2; // block transactions, in canonical order
+}
```
dshulyak, line 16: i think this is also not so useful, would be better to report block hash and provide List/Stream apis for rewards and transactions

poszu (approved):

spacemesh/v2alpha1/layer.proto
```diff
@@ -0,0 +1,49 @@
...
+message Layer {
+  oneof versioned {
+    BlockV1 v1 = 1;
+  }
+}
+
+message LayerV1 {
+  uint32 number = 1; // layer number - not hash - layer content may change
+  enum LayerStatus {
+    LAYER_STATUS_APPLIED = 0; // applied by hare
```
lrettig, line 16: Every protobuf enum should have an `_UNSPECIFIED` element. There needs to be a way to distinguish whether or not this has been explicitly set, and it's the only way to do that. I'm fine with the names `APPLIED` and `VERIFIED` for the other statuses.

spacemesh/v2alpha1/layer.proto
lrettig: What's the difference between state hash and "cumulative state hash" here?

spacemesh/v2alpha1/layer.proto
```diff
@@ -0,0 +1,50 @@
...
+  enum LayerStatus {
+    LAYER_STATUS_UNSPECIFIED = 0;
+    LAYER_STATUS_APPLIED = 1; // applied by hare
+    LAYER_STATUS_VERIFIED = 2; // verified by tortoise
+  }
+  LayerStatus status = 2;
+  string consensus_hash = 3;
+  bytes state_hash = 4; // fingerprint of the computed state at the layer
+  bytes cumulative_state_hash = 5; // cumulative fingerprint that uniquely identifies state since genesis
+  Block blocks = 6; // layer's blocks
```
lrettig, line 24: This is confusing. The plural suggests there's more than one, but `Block` is only a single block. See also #287.

lrettig (changes requested):

spacemesh/v2alpha1/layer.proto
kacpersaw: Changed

spacemesh/v2alpha1/layer.proto
dshulyak: we don't do cumulative at the moment. the state hash is a hash of the changed accounts in a given layer.

https://github.com/spacemeshos/go-spacemesh/issues/5677

lrettig:
@kacpersaw just a couple of clarifications from #287:

Can you confirm that we intend API v2 to _only_ return the canonical block for a layer, same as API v1? (https://github.com/spacemeshos/api/issues/287#issuecomment-2060005756)

How do we intend the API to reveal tx status and validity? Is the plan to add this later? (https://github.com/spacemeshos/api/issues/287#issuecomment-2061943758)

How do we intend the API to report reorg events?

Thanks.

kacpersaw:
@lrettig 
> Can you confirm that we intend API v2 to _only_ return the canonical block for a layer, same as API v1? ([#287 (comment)](https://github.com/spacemeshos/api/issues/287#issuecomment-2060005756))

Yes, only applied block will be returned.

> How do we intend the API to reveal tx status and validity? Is the plan to add this later? ([#287 (comment)](https://github.com/spacemeshos/api/issues/287#issuecomment-2061943758))

This PR is about LayerService, we can discuss this in TransactionService PR when it'll be created.

> How do we intend the API to report reorg events?

How do you want it to look like?

lrettig:
> > How do we intend the API to report reorg events?
> 
> How do you want it to look like?

There was some discussion of this in #38 but it was inconclusive. I think as a next step it would be good to see how APIs for other projects handle this case. For the purposes of this PR let's just do something simple and reasonable; we can continue the discussion in #38 to decide how we want to handle this more correctly and completely (ideally, API v2 should include this).

I don't see any issue with `LayerRequest`. For `LayerStream` maybe we just terminate the stream if we detect a reorg and then, for now, require the client to check on its end if there's been a reorg. As for how to do that, it's non trivial since https://github.com/spacemeshos/api/pull/319#discussion_r1572023386. In other words we still need a quick, easy way for a client to detect if a reorg has occurred.

I don't think this issue should block merging of this PR. Let's sort this out in #38.

lrettig (approved):

PR #320, opened 2024-03-20
Add PENDING post state

xearl4:
Currently, the PostStates endpoint reports the post state for managed identities to be either IDLE or PROVING. IDLE is for identities which are currently not scheduled to run a PoST proof. PROVING is for identities which should currently run a PoST proof.

However, PROVING currently does not help operators discern between identities for which a post-service is connected and those for which no post-service is connected.

This patch adds a third state, PENDING meant to indicate "is proving, but no post-service connected" vs PROVING as "is proving and already has a post-service connected". This allows operators to view PENDING as an indication that a post-service still needs to be connected.

This change will be followed up by a correlated change to go-spacemesh, to report the new state.


xearl4:
I now think it's better splitting this in two: one state indicating IDLE/PROVING, and a second state showing if a post service is DISCONNECTED/CONNECTED. Hence: closing.

PR #326, opened 2024-04-16
Trim down v2alpha1.ActivationV1

poszu:
We want to trim down the ATX domain type in the node to contain only the necessary fields. We will remove fields needed only for new ATX verification like POST and sequence number.  Additionally, the ATX weight will be calculated differently with the ATX merge (weight will be the sum of merged identities). Hence replaced `base_tick_height`, `tick_count`, and `num_units` with `weight` and `height`.

@kacpersaw are any of the removed fields needed in the explorer?


kacpersaw (approved):
LGTM

PR #329, opened 2024-04-30
v2alpha1: Add TransactionService

kacpersaw:
```
service TransactionService {
  rpc List(TransactionRequest) returns (TransactionList);
  rpc ParseTransaction(ParseTransactionRequest) returns (ParseTransactionResponse);
  rpc SubmitTransaction(SubmitTransactionRequest) returns (SubmitTransactionResponse);
  rpc EstimateGas(EstimateGasRequest) returns (EstimateGasResponse);
}
```

```
service TransactionStreamService {
  rpc Stream(TransactionStreamRequest) returns (stream Transaction);
}
```


spacemesh/v2alpha1/tx.proto
```diff
@@ -59,4 +59,106 @@ message TransactionV1 {
...
+    TRANSACTION_STATE_MEMPOOL = 4; // in mempool but not on the mesh yet
+    TRANSACTION_STATE_MESH = 5; // submitted to the mesh
+    TRANSACTION_STATE_PROCESSED = 6; // processed by STF; check Receipt for success or failure
+  }
+  TransactionState state = 2;
+}
+
+message TransactionResult {
+  enum Status {
+    SUCCESS = 0;
```
lrettig, line 83: please add `UNSPECIFIED` as zero, and let's use the ordinary enum naming syntax, e.g., `TRANSACTION_STATUS_UNSPECIFIED`

spacemesh/v2alpha1/tx.proto
```diff
@@ -59,4 +59,106 @@ message TransactionV1 {
...
+  string message = 3;
+  uint64 gas_consumed = 4; // in units of gas
+  uint64 fee = 5; // in smidge
+  bytes block = 6;
+  uint32 layer = 7;
+  repeated string touched_addresses = 8;
+}
+
+message TransactionStreamRequest {
+  uint32 start_layer = 1;
```
lrettig, line 99: are `start_layer` and/or `end_layer` optional? would be good to add a comment here to that effect (or can we indicate in the protobuf definition that they're optional?)

spacemesh/v2alpha1/tx.proto
```diff
@@ -59,4 +59,106 @@ message TransactionV1 {
...
+  uint32 end_layer = 2;
+  repeated bytes id = 3;
```
lrettig, line 101:
```suggestion
  repeated bytes txid = 3;
```

spacemesh/v2alpha1/tx.proto
```diff
@@ -59,4 +59,106 @@ message TransactionV1 {
...
+  string address = 4;
+  bool watch = 5;
```
lrettig, line 103: What does this mean? Can we add an explanatory comment?

spacemesh/v2alpha1/tx.proto
```diff
@@ -59,4 +59,106 @@ message TransactionV1 {
...
+}
+
+service TransactionStreamService {
+  rpc Stream(TransactionStreamRequest) returns (stream Transaction);
+}
+
+message TransactionRequest {
+  uint32 start_layer = 1;
+  uint32 end_layer = 2;
+  repeated bytes id = 3;
```
lrettig, line 113:
```suggestion
  repeated bytes txid = 3;
```

spacemesh/v2alpha1/tx.proto
```diff
@@ -59,4 +59,106 @@ message TransactionV1 {
...
+  string address = 4;
+  bool watch = 5;
+}
+
+service TransactionStreamService {
+  rpc Stream(TransactionStreamRequest) returns (stream Transaction);
+}
+
+message TransactionRequest {
+  uint32 start_layer = 1;
```
lrettig, line 111: same comments as above, are start and end layer optional?

spacemesh/v2alpha1/tx.proto
```diff
@@ -59,4 +59,106 @@ message TransactionV1 {
...
+  uint32 end_layer = 2;
+  repeated bytes id = 3;
+  string address = 4;
```
lrettig, line 114: What is `address` here? If it means principal we should rename to `principal`. Or does it mean any tx that touches an account? Let's also please add an explanatory comment.

lrettig (changes requested):

spacemesh/v2alpha1/tx.proto
```diff
@@ -59,4 +59,106 @@ message TransactionV1 {
 
 message Nonce {
   uint64 counter = 1;
+}
+
+// TransactionState is the "journey" of a tx from mempool to block inclusion to
+// mesh to STF processing. To know whether or not the tx actually succeeded,
+// and its side effects, check the Receipt in the GlobalStateService.
```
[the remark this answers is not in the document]
kacpersaw: Fixed comment

spacemesh/v2alpha1/tx.proto
kacpersaw: Fixed

spacemesh/v2alpha1/tx.proto
kacpersaw: Added comments

spacemesh/v2alpha1/tx.proto
kacpersaw: Added comments

spacemesh/v2alpha1/tx.proto
kacpersaw: Added comments and set to optional

spacemesh/v2alpha1/tx.proto
lrettig: I think we need to be more specific than "involved in." What does that mean precisely? Right now we could limit it to principal and recipient of a send, plus maybe spawn destination. In the future if we add more functionality at L1 this might get more complicated, or even be infeasible (e.g., there's basically no way for an Ethereum node to know which accounts were involved in a historical ERC-20 transaction, and it would be the same for some account state update that happened inside a tx at L2 in our case).

My inclination is to remove this for now. Let's just allow a request using `principal` since it should be easier to index on that. Maybe we can allow `recipient` of a coin send as well.

Do we know how we're going to implement this?

spacemesh/v2alpha1/tx.proto
```diff
@@ -96,26 +97,26 @@ message TransactionResult {
 }
 
 message TransactionStreamRequest {
-  uint32 start_layer = 1;
-  uint32 end_layer = 2;
-  repeated bytes id = 3;
-  string address = 4;
-  bool watch = 5;
+  uint32 start_layer = 1; // Starting layer for transaction streaming.
```
lrettig, line 100: Optional or required?

spacemesh/v2alpha1/tx.proto
```diff
@@ -96,26 +97,26 @@ message TransactionResult {
...
+  uint32 end_layer = 2; // Ending layer for transaction streaming.
+  repeated bytes txid = 3; // Filter: specific transaction IDs.
+  string address = 4; // Filter: address involved in transaction.
+  bool watch = 5; // Continue watching for new transactions.
```
lrettig, line 104: I don't understand this. By definition, using a stream means you're subscribing to new txs aren't you? What would happen if this were set to false?

spacemesh/v2alpha1/tx.proto
```diff
@@ -96,26 +97,26 @@ message TransactionResult {
...
-  string address = 4;
-  uint64 offset = 5; // adjusts the starting point for data
-  uint64 limit = 6; // specifies max number of items to fetch
-  bool include_state = 7;
-  bool include_result = 8;
+  optional uint32 start_layer = 1; // Starting layer for transactions.
+  optional uint32 end_layer = 2; // Ending layer for transactions.
+  repeated bytes txid = 3; // Filter: specific transaction IDs to filter.
+  optional string address = 4; // Filter: address involved in transaction.
+  uint64 offset = 5; // Adjusts the starting point for data retrieval.
```
lrettig, line 116: Required or optional?

lrettig (commented):
thanks for the changes!

spacemesh/v2alpha1/tx.proto
kacpersaw: start_layer will be required and end_layer optional

spacemesh/v2alpha1/tx.proto
kacpersaw: Stream is a private service - you can "stream" transactions for specific range of layers. Which will give you all transactions without limit/offset.

spacemesh/v2alpha1/tx.proto
kacpersaw: offset and limit is always required - just like in rewards, atxs, layers etc

spacemesh/v2alpha1/tx.proto
poszu: @lrettig 
>  and let's use the ordinary enum naming syntax, e.g., TRANSACTION_STATUS_UNSPECIFIED

Why use this prefix? It leads to auto-generated constants with redundant info in their name like: `TransactionState_TRANSACTION_STATE_UNSPECIFIED` instead of `TransactionState_UNSPECIFIED`.

spacemesh/v2alpha1/tx.proto
poszu: @lrettig 
> By definition, using a stream means you're subscribing to new txs aren't you?

It's not necessary. A server-streaming response differs from a unary response in that it doesn't return all elements at once but "streams" them one by one. [The GRPC spec](https://grpc.io/docs/what-is-grpc/core-concepts/#server-streaming-rpc) says the server can close the call when it finishes. In our case, when `watch=false`, it ends after streaming all TXs from the DB.

lrettig: Thanks guys, this makes sense

lrettig (approved):
lgtm!

spacemesh/v2alpha1/tx.proto
fasmat: Regarding `UNSPECIFIED` its part of the best practices for protobuf: https://protobuf.dev/programming-guides/dos-donts/#unspecified-enum

It is also recommended to prefix enums that belong together with the string: https://protobuf.dev/programming-guides/style/#enums I haven't found an explanation but I assume it is to avoid name collisions in some protobuf generators or assigning one enum value to a different enum (in languages that have no strict typing like Python or JS).

PR #339, opened 2024-06-03
v2alpha1: generate openapi v2 spec

kacpersaw:
Generate openapi spec using [protoc-gen-openapiv2](https://github.com/grpc-ecosystem/grpc-gateway/tree/main/protoc-gen-openapiv2).

`google.api.api_visibility.restriction` was added to all services to separate v1 from v2.

Example:
```
service ActivationService {
    option (google.api.api_visibility).restriction = "V1";
```

protoc-gen-openapiv2 generates openapi models for ALL messages which are in proto files. To remove unused models,  go-swagger dep was added to clean-up generated openapi json file.


fasmat (commented):
I think `release/openapi/api.swagger.json` should not be committed to the repository. It contains the same definitions as the `.proto` files again, but in JSON / Swagger and could be generated from the proto files by a user of the repository.

fasmat (commented):
I would prefer not having the swagger file in the repo, but it seems to be automatically checked to be in line with the proto files when running `make check` so I'm OK with it.

fasmat (approved):

PR #345, opened 2024-06-14
Swagger-UI for public API

andres-spacemesh:
Swagger json must be palced on api\release\openapi\swagger\src\


dsmello (approved):

fasmat:
The swagger api definition file is generated - so you need to update the `Makefile`/`buf.gen.yml` too or changes to the api won't be automatically updated 😉

fasmat:
@dsmello @andres-spacemesh

PR #351, opened 2024-06-19
v2alpha1: Add sort_order field to List requests

kacpersaw:
Add sort_order field (enum ASC/DESC) to tx, layer and reward List requests


spacemesh/v2alpha1/layer.proto
```diff
@@ -24,6 +25,7 @@ message LayerRequest {
   uint32 end_layer = 2;
   uint64 offset = 3; // adjusts the starting point for data
   uint64 limit = 4; // specifies max number of items to fetch
+  optional SortOrder sort_order = 5; // specifies the sort order (default is ASC)
```
poszu, line 28: Do you need the `optional` keyword if the default is the 0-value enum variant anyway (ASC)? Btw, I'd expect that the default (when not set) is to not sort at all. Wdyt?

spacemesh/v2alpha1/v2alpha1.proto
```diff
@@ -13,4 +13,9 @@ option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger) = {
...
     };
   };
-};
\ No newline at end of file
+};
+
+enum SortOrder {
+  ASC = 0;
+  DESC = 1;
+}
```
poszu, line 21: Two questions:
- what about "don't sort" variant? It could save some cycles when sorting is not needed.
- does the default need to be "sorted"?

kacpersaw: We can add third option which is "don't sort" but the current implementation sorts in ascending order, and I don't want to break existing integrations for others.

poszu (approved):

PR #355, opened 2024-06-21
v2alpha1: Add NumUnits to Activation and extend request filters

kacpersaw:
* Add NumUnits to Activation message
* Change name of NodeId filter to SmesherId.
* Modify Id to allow multiple entries.


brusherru (approved):

spacemesh/v2alpha1/activation.proto
```diff
@@ -13,6 +13,7 @@ message Activation {
   string coinbase = 6;
   uint64 weight = 7;
   uint64 height = 8;
+  uint32 num_units = 9; // number of PoST data commitment units
```
poszu, line 16: NIT:
```suggestion
  uint32 units = 9; // number of effective PoST data commitment units
```

This value might be smaller than the actual number of units in the ATX because we take the minimum of previous ATX and the current ATX. So somebody grewing their space and publishing ATXs with increasing size gets:
1. units = 4 SU, effective units = 4 SU
2. units = 8 SU, effective units = 4 SU
3. units = 8 SU, effective units = 8 SU

If you need the exact number of units (as it results from the commited PoST size in this ATX), we would need to read it from the ATX blob.

PR #358, opened 2024-06-26
v2alpha1: Remove total field from list request messages

kacpersaw:
Remove the total field from request messages because they are too heavy for the node to handle.


brusherru (approved):
👌

PR #359, opened 2024-06-27
Fix sync with go-sm api version

andres-spacemesh:


andres-spacemesh:
Moved to go-sm https://github.com/spacemeshos/go-spacemesh/pull/6088

fasmat (approved):

PR #362, opened 2024-07-22
v2alpha1: Add malfeasance API

kacpersaw:
```
message MalfeasanceProof {
    bytes smesher = 1;
    uint32 layer = 2;
    enum MalfeasanceType {
        MALFEASANCE_UNSPECIFIED = 0;
        MALFEASANCE_ATX = 1;
        MALFEASANCE_BALLOT = 2;
        MALFEASANCE_HARE = 3;
        MALFEASANCE_POST_INDEX = 4;
        MALFEASANCE_INCORRECT_PREV_ATX = 5;
        MALFEASANCE_DOUBLE_MARRY = 6;
    }
    MalfeasanceType kind = 3;
    string debug_info = 4;
    bytes proof = 5;
}
```

Add malfeasance services to v2alpha1


kacpersaw:
https://github.com/spacemeshos/go-spacemesh/issues/6150

fasmat:
This API will change a lot with changes in the near future:

- Malfeasance proofs will not have a `layer` identifier any more but instead a version identifier.
- There are now 3 domains for each proof:
  - Activation
  - Ballots
  - Hare
- Each domain can have any amount of sub-types of proofs. For Ballots and Hare at the moment there will only be a double publish proof, activation will have at least 5 sub-types.
  - The proof itself is wrapped in layers - the actual proof, wrapped in a domain-specific layer, wrapped in a general layer. The bytes returned however should in my opinion only be the concrete proof and not the wrapping layers.

For reference see also this these tickets:

- https://github.com/spacemeshos/pm/issues/331
- https://github.com/spacemeshos/go-spacemesh/issues/6143
- https://github.com/spacemeshos/go-spacemesh/issues/6140
- https://github.com/spacemeshos/go-spacemesh/issues/6141
- ... to be continued (I'm still creating issues 🙂 )

spacemesh/v2alpha1/malfeasance.proto
```diff
@@ -0,0 +1,54 @@
...
+
+    // Properties of the Malfeasance proof, different for every type of proof
+    map<string, string> properties = 4;
+}
+
+message MalfeasanceRequest {
+    repeated bytes smesher_id = 1; // list of smesher ids to fetch
+    uint64 offset = 2; // adjusts the starting point for data
+    uint64 limit = 3; // specifies max number of items to fetch
+}
```
poszu, line 33: What's the purpose of the `offset` and `limit` fields? If N `smesher_id`s are requested, I'd expect it to return from 0 to N proofs.

spacemesh/v2alpha1/malfeasance.proto
```diff
@@ -0,0 +1,54 @@
...
+
+message MalfeasanceList {
+    repeated MalfeasanceProof malfeasances = 1;
```
poszu, line 36: NIT, how about
```suggestion
    repeated MalfeasanceProof proofs = 1;
```

spacemesh/v2alpha1/malfeasance.proto
fasmat: It limits the number of proofs that are returned. `smesher_id` can be empty, then all known proofs are returned, but limited to `limit` elements, starting at `offset` of all proofs. So if 1000 proofs are available you can fetch them in chunks:

- request 1: offset 0, limit 100
- request 2: offset 100, limit 100
- etc.

We do the same in other places of the API (e.g. for ATXs).

poszu: Then the comment `// list of smesher ids to fetch` is incomplete. It should also say what you explained above. :)

fasmat: Updated

poszu (approved):

PR #369, opened 2024-08-19
API to provide info about poet registrations

ConvallariaMaj:
Related to issue https://github.com/spacemeshos/go-spacemesh/issues/6174

Connected PR:
https://github.com/spacemeshos/go-spacemesh/pull/6266


spacemesh/v2alpha1/smeshing_identities.proto
```diff
@@ -0,0 +1,56 @@
...
+message PoetServicesRequest {}
+
+enum IdentityStatus {
+  STATUS_UNSPECIFIED = 0;
+  STATUS_IS_SYNCING = 1;
+  STATUS_WAIT_FOR_POET_ROUND_START = 2;
+  STATUS_WAIT_FOR_POET_ROUND_END = 3;
+  STATUS_FETCHING_PROOFS = 4;
+  STATUS_POST_PROVING = 5;
+}
```
poszu, line 18: The variants don't need to be prepended, it results in awkward variable names like `RegistrationStatus_STATUS_FAILED_REG`.
```suggestion
enum IdentityStatus {
  UNSPECIFIED = 0;
  IS_SYNCING = 1;
  WAIT_FOR_POET_ROUND_START = 2;
  WAIT_FOR_POET_ROUND_END = 3;
  FETCHING_PROOFS = 4;
  POST_PROVING = 5;
}
```

spacemesh/v2alpha1/smeshing_identities.proto
```diff
@@ -0,0 +1,56 @@
...
+message PoetServicesResponse {
+  message Identity {
+    message PoetInfo {
+      string url = 1;
+      google.protobuf.Timestamp poet_round_end = 2;
+      RegistrationStatus registration_status = 3;
+      string warning = 4;
+    }
+
+    string smesher_id_hex = 1;
```
poszu, line 38: In all other APIs we use `bytes` for the smesher ID. IMHO we should stick to the same format.

spacemesh/v2alpha1/smeshing_identities.proto
```diff
@@ -0,0 +1,56 @@
...
+service SmeshingIdentitiesService {
+  option (google.api.api_visibility).restriction = "V2";
+
+  // Returns set of configured poet addresses and poets addresses from registrations, if given,
+  // and warning in case, if there are registrations with poets, which are not in configured poets set.
+  rpc PoetServices (PoetServicesRequest) returns (PoetServicesResponse){
+    option (google.api.http) = {
+      post: "/v2/smeshingidentities/poetservices"
+      body: "*"
+    };
```
kacpersaw, line 54: Other APIs use autogenerated endpoints; I think this can be removed.

spacemesh/v2alpha1/smeshing_identities.proto
fasmat: The protobuf style guide recommends prefixing enums: https://protobuf.dev/programming-guides/style/#enums it leads to some awkward variable names, but without it it might break other code generators 🤷

fasmat: @poszu how should we handle this? Go for cleaner names for us or use the recommendation by protobuf style guide?

ConvallariaMaj: I vote for cleaner names.

poszu: @fasmat The guide recommends prefixing instead of enclosing enum in a `Message` because some languages don't support putting an enum in a struct. But here, the enum is at the root level, it's not inside another Message so I think it's fine.

jellonek (approved):

kacpersaw:
This PR was used as part of the work to create https://github.com/spacemeshos/go-spacemesh/pull/6452.

PR #382, opened 2024-10-09
Introduce events about communication with PoET

jellonek:
First part of https://github.com/spacemeshos/go-spacemesh/issues/5565


spacemesh/v1/admin_types.proto
```diff
@@ -131,6 +135,29 @@ message EventMalfeasance {
 
 message EventStreamRequest {}
 
+message EventPoetClientCreated {
+  string url = 1;
+}
```
fasmat, line 140: I know this is part of the issue but this seems very technical and has little meaning to a user/operator: It only says that the node created an object internally with which it could connect to a PoET if it needs to and only happens during startup of the node (I think).

Maybe instead of removing we could also rename this event to something less technical: "`url` added to list of PoETs" or something similar.

Wdyt?

fasmat: The argument for not having this event is: if the config entry with the URL would be invalid or at the wrong place of the config, the node would refuse to start.

So if as a user/operator I add a new poet to the config and the node starts successfully the client for it was created.

jellonek: With my current changes the whole sequence of evens looks like:
```
Event_PoetClientCreated
Event_Beacon
Event_InitStart
Event_InitComplete
Event_PostServiceStarted
Event_PostStart
Event_PostComplete
Event_PoetWaitRound
Event_RegisteredInPoet
Event_PoetWaitProof
Event_PoetProofReceived
Event_BestPoetProofSelected
Event_PostStart
Event_PostComplete
Event_AtxPublished
```
(taken from node/node_test.go). At the moment all that looks "quite technical" :)

dunno if info about client creation is useful. More important is info about registration.

I can adapt that to any form you will suggest, but i need a bit more precise an input from someone who will be using that. 
It's right moment to tune all event names. E.g. maybe instead of `PoetProofReceived` there should `ProofDownloadedFromPoet`?

Technically event about poet client creation is emitted right after successful client creation. If it has to be renamed in proposed way, imo that will also require moving it in the node code to the place where poet service is added to poetClients list on node/node.go, in `initServices` method?

fasmat: `Event_Beacon` is arguably also a technical detail that doesn't need to be reported, but the others make sense to tell a user about:

- `InitStart`: Post initialization started, the node now creates post data for the identity it uses
- `InitComplete`: Post initialization completed, the node is now ready to participate in consensus
- `PostServiceStarted`: can be argued if needed or not, post is now handled by a dedicated process that can be run "supervised" (i.e. the node starts and manages that process) or manually by the user. If a user wants to manage multiple identities they have to start one post service per identity. I believe this event is still useful in communicating to the user that the node manages an identity and started post (service) for it.
- `PostStart`: The node has started generating a post proof (in your test most likely the initial post for the identity it manages)
- `PostComplete`: The node has completed generating a post proof

All of these can happen multiple times (once per configured PoET):
- `PoetWaitRound`: The node now wants to register at a PoET and waits until a registration window opens
- `PoetWaitProof`: The node registered at a PoET and is now waiting to receive a proof from it
- `PoetProofReceived`: The node successfully downloaded a proof from a PoET

Then the node does another post and publishes it's first ATX, after that events will be poet events, post events, atx publish, poet events, post events, atx publish, aso.

Now describing it: `RegisteredInPoet ` and `PoetWaitProof` are kinda saying the same thing,

fasmat: I believe `Event_PoetClientCreated` can be dropped completely unless @poszu objects.

poszu: Sure, it's not necessary. AFAIR, the idea was to give the user a confirmation about the poets that the given node will use.

spacemesh/v1/admin_types.proto
```diff
@@ -42,6 +42,11 @@ message Event {
     EventMalfeasance malfeasance = 15;
     EventPostServiceStarted post_service_started = 16;
     EventPostServiceStopped post_service_stopped = 17;
+    EventWaitingForPoETRegistrationWindow waiting_for_poet_registration_window = 18;
+    EventProofDownloadedFromPoet proof_downloaded_from_poet = 19;
+    EventRegisteredInPoet registered_in_poet = 20;
+    EventBestProofSelected best_proof_selected = 21;
+	EventWaitingForPoETRoundEnd waiting_for_poet_round_end = 22;
```
fasmat, line 49: NIT: Indentation is different from other lines 🙂

spacemesh/v1/admin_types.proto
```diff
@@ -131,6 +156,25 @@ message EventMalfeasance {
 
 message EventStreamRequest {}
 
+message EventRegisteredInPoet {
+  string url = 1;
+  bytes challenge = 2;
```
fasmat, line 161: I think this doesn't need to be part of the event. It's the hash of the header of the next ATX (not the hash/ID of the ATX), so not really useful to a user. However I think this event should include a `bytes smesher` field to know which identity was registered in PoET.

jellonek: That part was of requirements in the main [task](https://github.com/spacemeshos/go-spacemesh/issues/5565):
> registered in poet (URL, challenge, round)

If @poszu will agree - I can quickly remove `challenge`.
I'll `smasher`.

spacemesh/v1/admin_types.proto
jellonek: I'm still fixing project settings and looks like i need to enable showing me whitechars :D
Fixed, thx!

spacemesh/v1/admin_types.proto
poszu: Sure, the `challenge` is not critical. Unfortunately I don't recall why it was added. We can always add it later upon a community request. 

>However I think this event should include a bytes smesher field to know which identity was registered in PoET.

Probably all poet events should have it to tell which smesher given event is for, same as many other events. wdyt?

spacemesh/v1/admin_types.proto
```diff
@@ -79,15 +84,35 @@ message EventPostComplete {
...
+  uint32 target = 2;
+  google.protobuf.Duration wait = 3;
+  google.protobuf.Timestamp until = 4;
+  bytes smesher = 5;
+}
+
+message EventWaitingForPoETRoundEnd {
   uint32 publish = 1;
   uint32 target = 2;
   google.protobuf.Duration wait = 3;
```
poszu, line 118: These two differ only in their name. Perhaps it would be better to keep the old one (less complication in the API).

jellonek: The whole point there was to do the rename.
Having that duplicated leaves "old" untouched and simpler to remove in the future.
Keeping only one of them there would make emitters code in node `events/events.go` less readable IMO.

spacemesh/v1/admin_types.proto
```diff
@@ -79,15 +84,35 @@ message EventPostComplete {
...
   google.protobuf.Timestamp until = 4;
   bytes smesher = 5;
 }
 
+message EventWaitingForPoETRegistrationWindow {
+  uint32 current = 1;
+  uint32 publish = 2;
+  google.protobuf.Duration wait = 3;
+  google.protobuf.Timestamp until = 4;
```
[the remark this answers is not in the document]
fasmat: Yeah `until` is more useful than `wait` for automation, and `wait` can be calculated from it 🙂

fasmat (approved):

spacemesh/v1/admin_types.proto
poszu: @jellonek 
>Keeping only one of them there would make emitters code in node events/events.go less readable IMO.

I don't understand. Do you mean the name `EventPoetWaitProof` is less readable than `EventWaitingForPoETRoundEnd`?

jellonek: It a bit more precisely describes what happening in that part of the code.

spacemesh/v1/admin_types.proto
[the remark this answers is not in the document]
jellonek: Done. Instead of `until` we have chosen name `round_end`.

spacemesh/v1/admin_types.proto
poszu: What about the perspective of the consumer of this event (the user)? The user knows the node should wait for poet _proof_.

PR #389, opened 2024-11-18
node-slit: add smeshing identities service PoC

kacpersaw:


spacemesh/v2alpha1/smeshing_identities.proto
```diff
@@ -0,0 +1,50 @@
...
+}
+
+message Identity {
+  bytes smesher_id = 1;
+  repeated IdentityStateInfo history = 2;
+}
+
+message IdentityStatesResponse {
+  repeated Identity identities = 3;
+}
```
poszu, line 44: How about using a [map](https://protobuf.dev/programming-guides/proto3/#maps)?
```suggestion
message Identity {
  repeated IdentityStateInfo history = 1;
}

message IdentityStatesResponse {
  map<string, Identity> identities = 1;
}
```

The downside is that a map key cannot be `bytes` so you would need to encode it (e.g in hex). But on the other side, it would safe space because you wouldn't need to repeat the smesher_id.

poszu (approved):

PR #395, opened 2024-11-29
node-split PoC: Update IdentityStateInfo and PoetInfoResponse

kacpersaw:
## Description

* Added `ErrorInfo`, `PoETRegistration`, and `SubmittedRegistrations` message types in `smeshing_identities.proto`.
* Updated `IdentityStateInfo` to include `error`, `poet_url`, and `submitted_registrations` as part of a `oneof` metadata field.
* Replaced `PoetInfo` with `PoetConfig` and changed `PoetInfoResponse` to use a repeated string for `poets` in `smeshing_identities.proto`.


spacemesh/v2alpha1/smeshing_identities.proto
```diff
@@ -30,11 +31,56 @@ enum IdentityState {
...
+  bytes challenge_hash = 1;
+  string address = 2;
+  string round_id = 3;
+  google.protobuf.Timestamp round_end = 4;
+}
+message PoetRegisteredState {
+  repeated PoETRegistration registrations = 1;
+}
+
+message WaitForPoetRoundEndState {}
```
poszu, line 51: Perhaps it could also include a timestamp of the round end?

poszu (approved):

spacemesh/v2alpha1/smeshing_identities.proto
kacpersaw: Done

PR #407, opened 2024-12-20
node-split poc: Add eligible state

kacpersaw:


spacemesh/v2alpha1/smeshing_identities.proto
```diff
@@ -72,6 +74,15 @@ message ProposalPublishedState {
   uint32 layer = 2;
 }
 
+message Eligibility {
+  uint32 layer = 1;
+  uint32 count = 2;
+}
+message Eligible {
+  uint32 epoch = 1;
```
poszu, line 82: Technically you don't need this - the epoch can be derived from the layer number

poszu (approved):

PR #423, opened 2025-01-30
Added definitions for versioning

pigmej:


poszu (changes requested):
You need to run `make build` to generate the Go and swagger files ;)

poszu (approved):

